PR #2, opened 2020-10-31
Dit back update

ciliakas:
Integrated the current back end configuration into current display
Income.cshtml.cs and Expenses.cshtml.cs showcase different ways of using it


src/DataBase/Models/Expenses.cs
```diff
@@ -15,6 +16,26 @@ public class Expenses
...
+        }
+
+        public void Edit(Entry newEntry)
+        {
+            Amount = newEntry.Amount;
+            Title = newEntry.Title;
+            Date = newEntry.Date;
+            IsMonthly = newEntry.IsMonthly;
+            Importance = newEntry.Importance;
+        }
```
Stokalas, line 38: Since I'm not very familiar with DB `Edit` is clear to me but that think without name is kinda interesting what is it?
I can see same thing with incomes, but no point commenting same thing.

src/DataManagement/EntryList.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+using System.Threading.Tasks;
+using ePiggyWeb.Utilities;
+
+namespace ePiggyWeb.DataManagement
+{
+    public class EntryList : List<Entry>
```
Stokalas, line 11: Why should this be used instead of simple Lists of entries? Just curious

src/DataManagement/EntryList.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            var sb = new StringBuilder();
+
+            foreach (var entry in this)
+            {
+                sb.Append(entry);
+                //sb.Append("\n");
+            }
+
+            return sb.ToString();
```
Stokalas, line 33: Where are you seeing this `ToString()` to be used?

src/Pages/Expenses.cshtml.cs
```diff
@@ -7,20 +7,12 @@ namespace ePiggyWeb.Pages
...
-            using var context = new DatabaseContext();
-            var expense = context.Expenses; // define query
-            foreach (var item in expense.Where(x => x.UserId == 0)) // query executed and data obtained from database
-            {
-                var newExpense = new DataEntry(item.Id, item.UserId, item.Amount, item.Title, item.Date, item.IsMonthly,
-                    item.Importance);
-                Expenses.Add(newExpense);
-            }
+            var dataManager = new DataManager();
+            Expenses = dataManager.Expenses.EntryList;
```
Stokalas, line 15: As I can understand these 2 lines give the same result as all deleted ones? If so it's very neat optimisation. Is it some sort related to that thing called dependency injection?

Stokalas (approved):
I can see that changes are done very well and seems logical. Just some parts are different than what we had so I left questions, mainly to have a better understanding of the code. Approving since no faults detected 🙂  (did not test, just read the code)

Stokalas:
And as I understand you have already improved what was in the Income/Expenses pages so no point to create pull request for that branch? Or am I wrong?

src/DataBase/Models/Expenses.cs
ciliakas: I'm sorry but I'm not sure I understand the question, could you rephrase please?

src/DataManagement/EntryList.cs
ciliakas: There were a few reasons on why I implemented this change:
- First this class contains EntryType variable, which saves a lot of code later on when calling methods and so on.
- If we ever need to add more specific requirements for a list containing Entries, this will make it easier not to have to create a new class and amend the code everywhere. 
- We can create override methods if we need to for the List class
- Debatable, but could improve readability and ease of writing code, instead of always writing List<Entry> we can just write EntryList
- I think that this might be useful later to implement Dependency Injection and other more difficult requirements

src/DataManagement/EntryList.cs
ciliakas: This is not a critically important method, but might be useful for testing purposes, also the ToString method in Entry showcases was useful to test out reflection :D

src/Pages/Expenses.cshtml.cs
ciliakas: To the first part of the question, yes, the DataManager represents the old Data class, but now I've moved the reading from database code into the constructor. Also I haven't made the pull request, but I also implemented Lazy initialization for the variables in DataManager, so now we will only get the data from database only when we actually call the lists we need.
To the second part, this still isn't dependency injection but I'm learning about it as we speak and will try to implement it soon, just need to wrap my head around it :D

ciliakas:
> 
> 
> And as I understand you have already improved what was in the Income/Expenses pages so no point to create pull request for that branch? Or am I wrong?

You're correct

src/DataBase/Models/Expenses.cs
Stokalas: I think I've got it. My question was what is this method `public Expenses(...` And I think it's constructor :) But at that time I was confused so...

ciliakas: Yeah it's one of the constructors 👍

PR #3, opened 2020-11-02
Display income table

ciliakas:
An front end update and part of back end port:
Front:
- Added displaying of Incomes and Expenses

Back:
- Ported some utility functions
- Ported most of Data Entries, but changed the structure a bit, should now take up less code and be easier to use ( though it might take some time to familiarise with the new code)


Stokalas (approved):
As I skimmed almost all of these files are moved from earlier project and have been already reviewed. So nothing much to say here, but Im glad that backend files are being gathered in new project.

ciliakas:
Since the code has been review on earlier requests, I will merged without further reviews.

PR #4, opened 2020-11-02
Goals, register pages

Stokalas:
**Additions:**
- Goals.cshtml
- Register.cshtml
- Some Log In code that doesnt seem to work.

### DIFF
--- a/src/Pages/Expenses.cshtml
+++ b/src/Pages/Expenses.cshtml
@@ -10,9 +10,9 @@
         <a class="btn btn-primary form-control">Add Expense</a>
     </div>
 </div>
-<div class="table-wrap col-12 border p-3 mt-3 mb-3">
+<div class="col-12 border p-3 mt-3 mb-3">
         <form method="post">
-            @if (Model.Expenses.Count() > 0)
+            @if (Model.Expenses.Any())
             {
                 <table class="table table-hover border-primary table-bordered">
                     <tr class="table-primary">
--- a/src/Pages/Goals.cshtml.cs
+++ b/src/Pages/Goals.cshtml.cs
@@ -1,8 +1,3 @@
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Threading.Tasks;
-using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.RazorPages;
 
 namespace ePiggyWeb.Pages
--- a/src/Pages/Goals.cshtml.cs
+++ b/src/Pages/Goals.cshtml.cs
@@ -1,0 +1,16 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.RazorPages;
+
+namespace ePiggyWeb.Pages
+{
+    public class GoalsModel : PageModel
+    {
+        public void OnGet()
+        {
+        }
+    }
+}
--- a/src/Pages/logIn.cshtml
+++ b/src/Pages/logIn.cshtml
@@ -1,7 +1,5 @@
 ﻿@page
 @model ePiggyWeb.Pages.logInModel
-@{
-}
 
 <div class="container login-container">
     <div class="login-form-2">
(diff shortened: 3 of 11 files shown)


src/Pages/Goals.cshtml.cs
```diff
@@ -0,0 +1,16 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
```
ciliakas: Using quite a few unnecessary dependancies here

src/Pages/logIn.cshtml
```diff
@@ -1,25 +1,25 @@
...
             <div class="form-group">
                 <input type="submit" class="btnSubmit" value="Login" />
             </div>
             <div class="form-group">
-
                 <a href="#" class="ForgetPwd" value="Login">Forget Password?</a>
             </div>
+            <div class="form-group">
+                <a asp-page="/Register" class="ForgetPwd" value="Login">New User?</a>
+            </div>
```
ciliakas: Just curious, how does this work? I know there are some issues at the moment but in general at least

src/Pages/logIn.cshtml.cs
```diff
@@ -1,7 +1,9 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading.Tasks;
```
ciliakas: Using a few unnecessary dependancies again, and also I believe that LogIn should start with an upper case letter to match our standard :)

src/wwwroot/css/styles.css
```diff
@@ -152,6 +152,7 @@ body {
 }
 
 .table-wrap {
-    height: 80px;
+    height: 50%;
     overflow-y: auto;
 }
+
```
ciliakas: Can' comment, but on line 31 in this file Visual Studio is telling there is an error, though it still compiles and runs fine

ciliakas (changes requested):
Everything looks fine, the front end looks good, but I'd like the logIn.cshtml to be renamed to match our naming standards.

src/Pages/logIn.cshtml
Stokalas: This is a basic html form, which allows to use `Post` method when some button (in this case Login button) is pressed.
interesting part is `asp-for`. It allows to save data from input fields into Properties that are inside `logIn.cshtml.cs` and then use them or pass them. It is interesting because it somewhat different on how _Javascript_ handles input fields using _node.js_, but at the same time somewhat similar, but that's not important. 
If you have any further questions don't hesitate to ask :)

src/wwwroot/css/styles.css
Stokalas: Since it is _CSS_ it will always compile, but in case of error it wont display something right, I'll take a look and try to fix it, thanks!

src/Pages/Goals.cshtml.cs
Stokalas: Did not pay attention to this file tbh, since Goals only have design and no functionality, **yet**, but still good notice 🙂

src/Pages/logIn.cshtml
ciliakas: Thanks for the explanation! The asp-for is actually what got me to ask the question in the first place

Stokalas:
**Since there were some additions I'll list them here:**
- You can now "_login_" and if succesful it will transfer to `index`, if not - error message  displayed.
- No more income/expenses table should be limited and have scrollbar (probably will implement this later)

src/Pages/Expenses.cshtml
```diff
@@ -10,9 +10,9 @@
         <a class="btn btn-primary form-control">Add Expense</a>
     </div>
 </div>
-<div class="table-wrap col-12 border p-3 mt-3 mb-3">
+<div class="col-12 border p-3 mt-3 mb-3">
         <form method="post">
-            @if (Model.Expenses.Count() > 0)
+            @if (Model.Expenses.Any())
```
ciliakas, line 15: Forgot to mention this before, good change👍

ciliakas (approved):
All issues addressed, approved

PR #5, opened 2020-11-03
Interface and dependancy injection update

ciliakas:
I know this looks like a monster commit, and it is a bit, but a lot of the stuff is pretty simple or just refactoring.
The files to look at are in the Data Management folder and Database updaters.
I've partly implemented dependency injection into our Entry and Goal logic. Instead of using concrete classes like Entry or Goal, most of the methods and constructors use interfaces like IEntry or IGoal, which should help a lot with code reuseability. This is a bit of a complex topic and I'm not very good at it yet, so if any of the code or logic is unclear please just ask.


src/DataBase/EntryDbUpdater.cs
```diff
@@ -103,32 +103,32 @@ public static bool RemoveEntryList(EntryList entryList)
...
-                    entry.Id = databaseEntry.Id;
+                    var dbEntry = new ExpenseModel(entry, userId);
+                    dbEntryList.Add(dbEntry);
+                    entry.Id = dbEntry.Id;
                 }
-                db.AddRange(databaseEntryList);
+                db.AddRange(dbEntryList);
             }
 
             db.SaveChanges();
```
Stokalas, line 138: As I understand most of what happened with db was some renaming of models?

src/DataManagement/Entries/Entry.cs
```diff
@@ -1,66 +1,99 @@
...
+
+        public Entry(int id, IGoal goal, DateTime date, bool recurring, int importance) 
+            : this(id, goal.UserId, goal.Title, goal.Amount, date, recurring, importance) { }
+
+        public Entry()
 		{
 			Id = 0;
             UserId = 0;
 			Amount = 0;
 			Title = "unnamed";
```
Stokalas, line 50: Probably just weird github indentation

src/DataManagement/Entries/EntryList.cs
```diff
@@ -1,20 +1,14 @@
...
 using ePiggyWeb.Utilities;
 
-namespace ePiggyWeb.DataManagement
+namespace ePiggyWeb.DataManagement.Entries
 {
-    public class EntryList : List<Entry>
+    public class EntryList : List<IEntry>, IEntryList
     {
         //EntryType is to save up on the amount of times we have to pass the EntryType in methods
         public EntryType EntryType { get; set; }
```
Stokalas: Can you explain that comment a bit more detailed?

src/DataManagement/Entries/EntryManager.cs
```diff
@@ -1,109 +1,108 @@
...
-using ePiggyWeb.DataBase.Models;
 using ePiggyWeb.Utilities;
-using Microsoft.EntityFrameworkCore;
+using IListExtension;
 
-namespace ePiggyWeb.DataManagement
+namespace ePiggyWeb.DataManagement.Entries
 {
-    public class EntryManager
+    public class EntryManager : IEntryManager
```
Stokalas, line 9: There's a bit of code in this class, but maybe you can give short description and general purpose of what it does?

src/DataManagement/Entries/IEntry.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using ePiggyWeb.DataManagement.Goals;
+
+namespace ePiggyWeb.DataManagement.Entries
+{
+    public interface IEntry : IGoal
```
Stokalas, line 6: Is it good practice to have Entry inherit from Goal? Since I know base properties are same, but if we changed everything inside of goal? Just wondering.

src/DataManagement/Goals/Goal.cs
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        private static readonly string ResourceDirectoryParsedGoal = Directory.GetParent(Environment.CurrentDirectory)
+                                                                         .Parent?.Parent?.FullName +
+                                                                     @"\resources\textData\parsedGoal.txt";
+        public void SetGoalFromWeb(string itemName)
+        {
+            try
+            {
+                Task.Run(() => InternetParser.ReadPriceFromCamel(itemName)).Wait();
```
Stokalas: Just noting that this line 61 is not correct as I remember and should be edited if better knowledge about thread already present.

src/Pages/Income.cshtml.cs
```diff
@@ -1,16 +1,16 @@
...
     public class IncomesModel : PageModel
     {
-        public EntryList Income { get; set; }
+        public IEntryList Income { get; set; }
         public void OnGet()
         {
-            var incomeManager = new EntryManager(new EntryList(EntryType.Income));
-            Income = incomeManager.EntryList;
+            var dataManager = new DataManager();
+            Income = dataManager.Income.EntryList;
```
Stokalas, line 13: Are these changes just because restructure/remaining of backend code or foes it change anything?

src/Pages/Index.cshtml
```diff
@@ -46,8 +46,8 @@
                 </div>
                 <div class="col-lg-4 feature-box">
                     <p><i class="fas fa-bullseye fa-4x icon-features"></i></p>
-                    <h3 class="header-features">Goals</h3>
-                    <p>Enter goal by yourself or just provide name and we will find price for you.</p>
+                    <h3 class="header-features">GoalModel</h3>
+                    <p>Enter goal by yourself or just provide name and we will find amount for you.</p>
```
Stokalas: Why exactly this was renamed to GoalModel? :D Because it seems like  Resharper's job 😆

src/Pages/Shared/_Layout.cshtml
```diff
@@ -44,7 +44,7 @@
                     <div class="dropdown-menu" aria-labelledby="navbarDropdown">
                         <a class="dropdown-item" href="#">Spending</a>
                         <a class="dropdown-item" href="#">Balance</a>
-                        <a class="dropdown-item" href="#">Incomes/Expenses</a>
+                        <a class="dropdown-item" href="#">IncomeModel/Expenses</a>
```
Stokalas: Same here as with GoalModel

src/Pages/Index.cshtml
```diff
@@ -34,7 +34,7 @@
                 <div class="col-lg-4 feature-box">
                     <p><i class="fas fa-magic fa-4x icon-features"></i></p>
                     <h3 class="header-features">Automated input</h3>
-                    <p>Incomes/expenses that occur every month are entered automatically.</p>
+                    <p>IncomeModel/expenses that occur every month are entered automatically.</p>
```
Stokalas: Weird renaming

src/Utilities/DataJSON.cs
```diff
@@ -0,0 +1,116 @@
...
+
+				file.Close();
+			}
+			catch (FileNotFoundException f)
+			{
+				ExceptionHandler.Log(f.ToString());
+			}
+		}
+	}
+}
```
Stokalas, line 116: Could you elaborate where do you see this being used? Because when I think about it, it's illogical to save inside server whole data and we're not using _Javascript_.

src/Utilities/QueueList.cs
```diff
@@ -0,0 +1,54 @@
+﻿namespace ePiggyWeb.Utilities
+{
+    public class QueueList<T>
```
Stokalas, line 3: This class is also questionable, it was only to fulfill requirements of the first part, but as I understand it is totally unnecessary to use it now

Stokalas (changes requested):
I've left quite a few comments. And it seems problem with big pull requests is that it is hard to stay concentrated and check a lot of code, so it would be better to do smaller ones (I understand that all changes related, but still).
Another thing is that I think when you answer me about certain part of code, how it works what it does, it would be nice to put same answer into code as a comment, because I believe possibility is high that I'll forget something and it would be really frustrating to go and search for answer in a merged pull request,
And last thing, I'm requesting changes, because that renaming in html i.e. `Income` -> `IncomeModel` seems fishy to me.

src/DataBase/EntryDbUpdater.cs
ciliakas: I renamed the models, because (at least to me) having classes named Goal and Goals was a bit unclear and I would get confused a lot :D I also deleted old migrations which weren't necessary

src/DataManagement/Entries/Entry.cs
ciliakas: Yeah, looks fine in Visual Studio

src/DataManagement/Entries/EntryList.cs
ciliakas: Yeah the comment is really unclear, what I meant was that instead of having to add and pass a paramater of EntryType to most of the methods that involve EntryLists, we just have it as a variable inside the EntryList.
Basically just to make methods easier to call

src/DataManagement/Entries/EntryManager.cs
ciliakas: So EntryManager (and by extension GoalManager) basically takeover the same functionality as old Data.cs class used to do.
EntryManager reflects all the change we do on a local list to the database as well. In other words, if we want to add a entry, we call the EntryManager.Add method, which will first update the database, and if the database was successfully updated, it will update the local list as well, and analogously for every other method in EntryManager.
Also EntryManager has the duty of reading the data from the database on it's creation.

src/DataManagement/Entries/IEntry.cs
ciliakas: That's a good point, and I'm really not sure on the best practice here. My reasoning was that at the core, these two types are very connected, and since we want to have the feature of converting a goal into expense, this kind of neatly connects the two.
At the same time, since these are just interfaces and not concrete classes, it wouldn't take too much time to implement this change if the need arose.
I'm personally leaning into keeping this as is, but If you think this should be amended I don't mind doing so

src/DataManagement/Goals/Goal.cs
ciliakas: Thanks for the reminder! I marked it down as a comment, and also commented out the whole section related to reading goals from web, since I don't have the knowledge to port it properly yet into our new project

src/Pages/Income.cshtml.cs
ciliakas: The previous lines 12 and 13 were just showcasing how EntryManager class works, in this version I changed it to how I imagined that we would be using the back end - everything through a single DataManager class.

src/Pages/Index.cshtml
ciliakas: Ooops

src/Pages/Index.cshtml
ciliakas: Yup, you're correct, when I was renaming database models I accidentally left the box for string renaming checked as well, so that's how all of these errors are here. I tried to catch them, but I as you can see I missed a few. Thanks a lot for pointing them out 👍

src/Utilities/DataJSON.cs
ciliakas: Good point, I just copied this over without thinking too much, do you think we should just scrap this class?

src/Utilities/QueueList.cs
ciliakas: I copied this over thinking we might need this for calculations, or to at least see how this was used in calculations to be able to move away from it more easily. Do you think I should just delete this class?

ciliakas (commented):
You're completely right on the size of the pull requests, this was really poorly done on my part. After this pull request, I really tried to keep my branches way smaller, so a pull request of this size should happen again.
I've will leave a few of these answers as comments in code and add a bit more explanation in general as comments.
And yeah, the name changes were a mistake on my part, will fix those as well.
Thanks for your in depth review, really helped me out with catching a lot of the issues 👍

src/Utilities/DataJSON.cs
Stokalas: Well it could stay for a time and if we wont inplement it, it can be removed

src/Utilities/QueueList.cs
Stokalas: Same as with one with Json I/O, it can stay for the time being

Stokalas (approved):
Now everything seems +/- fine. I'm not sure how to use some parts of code yet, but I think everything will be allright after reading it a few times more :D

ciliakas:
> 
> 
> Now everything seems +/- fine. I'm not sure how to use some parts of code yet, but I think everything will be allright after reading it a few times more :D

Yeah there's quite a few change and if you ever unclear on how it works just ask

PR #6, opened 2020-11-04
Filtering Port

ciliakas:
A simplified version of our previous filtering and totals calculating system, and should be able to do almost  everything our previous system did.
Entry.cshtml.cs is an example of how to use it

### DIFF
--- a/src/DataManagement/Entries/EntryList.cs
+++ b/src/DataManagement/Entries/EntryList.cs
@@ -29,34 +29,38 @@ namespace ePiggyWeb.DataManagement.Entries
 
         public IEntryList GetBy(DateTime dateTime)
         {
-            throw new NotImplementedException();
+            return new EntryList(EntryType, this.Where(x => x.Date.Year == dateTime.Year && x.Date.Month == dateTime.Month).ToList());
         }
 
-        public IEntryList GetBy(DateTime @from, DateTime to)
+        public IEntryList GetBy(DateTime from, DateTime to)
         {
-            throw new NotImplementedException();
-        }
-
-        public IEntryList GetBy(Importance importance, DateTime @from, DateTime to)
-        {
-            throw new NotImplementedException();
-        }
-
-        public IEntryList GetBy(Importance importance, DateTime dateTime)
-        {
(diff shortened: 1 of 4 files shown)


src/Pages/Entry.cshtml.cs
```diff
@@ -32,7 +35,8 @@ public void OnGet()
             IGoal goal2 = new Goal();
             //IEntry goal3 = new Goal(); -- thats illegla
 
-            ViewData["EntryList"] = entry1.ToString();
+            //This should be ill eagle, don't know why I didn't think of this before
+            ViewData["EntryList"] = dataManager.Income.EntryList.GetBy(Importance.Low).GetBy(DateTime.Today).GetSum();
```
Stokalas, line 39: This seems very neat

src/DataManagement/Entries/EntryList.cs
```diff
@@ -8,12 +10,58 @@ public class EntryList : List<IEntry>, IEntryList
...
+            EntryType = entryType;
+            if (entryList is null) return;
+            AddRange(entryList);
+        }
+        
+        public IEntryList GetBy(Importance importance)
+        {
+            return new EntryList(EntryType, this.Where(x => x.Importance == (int)importance).ToList());
+        }
+
```
Stokalas, line 29: Are all these methods avaible for Income/Expenses or for the Goals as well? Not necessarily that we would use it for goals, just interesting.

Stokalas (approved):
Since this one is a lot shorter, not much to say. I really like how you simplified all those filtering/total methods into so few. Really well done. And as always I've left few questions.

_Approving since no flaws found_

src/DataManagement/Entries/EntryList.cs
ciliakas: I've just added to Incomes and Expenses for now at least

Stokalas (commented):
Just a question, so if I want to get entries only from this month I should use ?

ciliakas:
> 
> 
> Just a question, so if I want to get entries only from this month I should use ?

GetBy(DateTime.Today)

PR #7, opened 2020-11-04
Income/Expenses Front update

Stokalas:
- Updated Income.cshtml and Expenses.cshtml to match somewhat final version of design.
- Some logic of parsing data from input and preparing to add, but since it crashes once Post called no further developement.
(expenses.cshtml.cs) 

Tried to keep Pull request as small as possible 🙂


src/Pages/Expenses.cshtml.cs
```diff
@@ -8,12 +10,41 @@ public class ExpensesModel : PageModel
...
+            {
+                Error = "Amount is not a number!";
+                return;
+            }
+
+            var parsedDate = Convert.ToDateTime(Date);
+            var parsedIsMonthly = Convert.ToBoolean(IsMonthly);
+            var parsedImportance = int.Parse(Importance);
+            var Entry = new Entry(Title, parsedAmount,parsedDate,parsedIsMonthly,parsedImportance);
+            //kodas prid�t � listus (b�t� gerai jei Arnas padaryt�)
```
ciliakas: To add an Entry, just call Expenses.EntryList.Add(Entry). It returns a boolean if it worked

ciliakas: As discussed, I'm not sure how to pass the DataManager object around, but for now if we just copy code in OnGet it seems to work, and we need to apply the fix I've told you about in the EntryDbUpdater

ciliakas (changes requested):
The front end itself looks neat and looks like has all the features that we will need, the code looks neat as well, just need to fix the glitch with adding and this can be approved :)

src/Pages/Expenses.cshtml.cs
Stokalas: Is this working in current files, or in the new not merged pull request, since this exact line doesn't seem to work ?

ciliakas: My mistake, should work with dataManager.Expenses.EntryList.Add(Entry)

src/Pages/EditEntry.cshtml
```diff
@@ -0,0 +1,79 @@
...
+    {
+        <input asp-for="Entry.Recurring" type="checkbox" />
+    }
+    <input type="submit" value="Save" class="btn btn-primary ml-5" />
+    <input asp-page-handler="Cancel" class="btn btn-danger ml-2" type="submit" value="Cancel">
+
+    <br />
+    <div class="container p-0 mt-5 mb-4">
+        <h4 class="text-info">Quotes by famous people</h4>
+    </div>
```
ciliakas: I know this is totally a minor thing, but if it's easy to do, would it be possible to make the ui not move when going from Income to editing? As in, the Income label stays in exactly the same spot, but all the input fields move just a tiny bit down, would be neat if they also stayed in the same place.

src/Pages/EditEntry.cshtml.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        /*If cancel pressed return to previous page*/
+        public void OnPostCancel()
+        {
+            Debug.WriteLine("\n\n\n\n"  + EntryTypeInt);
+            if (EntryTypeInt == 1)
+            {
+                Response.Redirect("/Income");
+            }
+            Response.Redirect("/Expenses");
```
ciliakas: This doesn't redirect as it should, because it always goes to expenses, and I think it's because there's no return statement at the end of the if statement, so both Redirect methods are called.

ciliakas (changes requested):
To start with, the add, edit and delete functions work as intended. However I noticed a few issues:
- When delete is clicked, the red error labels that should only appear when tried to add
- The redirecting from EditEntry always redirects to expenses (while I'm not allowed when adding)

ciliakas (changes requested):
All issues addressed except that edit form still allows blank names

ciliakas (approved):
All issues resolved, very well done

PR #8, opened 2020-11-04
added removing and editing by id

ciliakas:
Added editing, removing and removing list by id


Stokalas (approved):
Very needed update!

Stokalas (approved):
Good update since it is much easier to use methods for DB with ID in front. Not sure about EntryManager, since it requires for object to be created, and that's what I want to avoid in front. Anyway maybe Dependency Injection will solve that. Well done.

PR #9, opened 2020-11-05
User Authorization minor update

ciliakas:
Changes:
- Registration now returns -1 if email is already in use and a user Id which is > 0 if successful
- Login now returns -1 if no such email exists, -2 if password is wrong, and a user Id which is > 0 if successful
- Made variable names match across the class, to be 'user'
Additions:
- Added a method to return user id by email, returns -1 if no such email is in use, and a user Id which is > 0 if successful

### DIFF
--- a/src/Authentication/UserAuth.cs
+++ b/src/Authentication/UserAuth.cs
@@ -38,23 +38,40 @@ namespace ePiggyWeb.Authentication
             return true;
         }
 
-        public static bool Login(string email, string pass)
+        public static int Login(string email, string pass)
         {
             using var db = new DatabaseContext();
             var userInfo = db.Users.FirstOrDefault(a => a.Email == email); //Find user and pass in db and check if matches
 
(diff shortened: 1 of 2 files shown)


src/Authentication/UserAuth.cs
```diff
@@ -34,43 +35,59 @@ public static bool Registration(string email, string pass)
...
-            return true;
+            return user.Id;
+        }
+
+        public static int GetUserIdByEmail(string email)
+        {
+            if (email is null)
+            {
+                return -1;
+            }
```
Stokalas, line 65: Nice optimisation, I thought why would we need this since if it is empty it won't be found in database, and in this case it wont even make connection to the DB.

Stokalas (approved):
Small change so nothing much to say, but helpful one since it makes saving user ID in cookie a lot easier.

Nice that you added comments

PR #10, opened 2020-11-05
Ported monthly updater

ciliakas:
Ported the monthly updater class. At the moment it is not integrated into Data Manager, so if need be should be used manually. Methods:
- Update Recurring: updates all the recurring entries from an Entry Manager, this is the method that we used to use before on every start up and after every addition or edition of an entry
- Update Entry: this is supposed to be called when we edit an entry an make it recurring
- Add Monthly Entry: this is supposed to be called when we add a recurring entry

### DIFF
--- a/src/DataManagement/RecurringUpdater.cs
+++ b/src/DataManagement/RecurringUpdater.cs
@@ -1,0 +1,125 @@
+﻿using System;
+using ePiggyWeb.DataBase;
+using ePiggyWeb.DataManagement.Entries;
+using ePiggyWeb.Utilities;
+
+namespace ePiggyWeb.DataManagement
+{
+    public static class RecurringUpdater
+    {
+        public static bool UpdateRecurring(IEntryManager entryManager)
+        {
+            foreach (var entry in entryManager.EntryList.GetBy(true))
+            {
+                var differenceInMonths = TimeManager.DifferenceInMonths(laterTime: DateTime.Today, earlierTime: entry.Date);
+                if (differenceInMonths > 0)
+                {
+                    if (!UpdateEntry(entryManager, entry, differenceInMonths))
+                    {
+                        return false;
+                    }
+                }
+            }
+            return true;
+        }
+
+        private static bool UpdateEntry(IEntryManager entryManager, IEntry entry, int months)
+        {
+            var nextMonth = entry.Date;
+            // one less since last one has to keep isMonthly = true;
+            var tempList = new EntryList(entryManager.EntryList.EntryType);
+            for (var i = 0; i < months - 1; i++)
+            {
+                //Adding new entry for each month according to date difference
+                nextMonth = TimeManager.MoveToNextMonth(dateTime: nextMonth);
+                var newEntry = new Entry(entry.Title, entry.Amount, nextMonth, false, entry.Importance);
+                tempList.Add(newEntry);
+            }
+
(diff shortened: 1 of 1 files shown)


src/DataManagement/RecurringUpdater.cs
```diff
@@ -0,0 +1,125 @@
...
+
+namespace ePiggyWeb.DataManagement
+{
+    public static class RecurringUpdater
+    {
+        public static bool UpdateRecurring(IEntryManager entryManager)
+        {
+            foreach (var entry in entryManager.EntryList.GetBy(true))
+            {
+                var differenceInMonths = TimeManager.DifferenceInMonths(laterTime: DateTime.Today, earlierTime: entry.Date);
```
Stokalas, line 14: As lecturer mentioned we can probably use just `laterDate - earlierDate`. Atleast I think

src/DataManagement/RecurringUpdater.cs
```diff
@@ -0,0 +1,125 @@
...
+using ePiggyWeb.DataManagement.Entries;
+using ePiggyWeb.Utilities;
+
+namespace ePiggyWeb.DataManagement
+{
+    public static class RecurringUpdater
+    {
+        public static bool UpdateRecurring(IEntryManager entryManager)
+        {
+            foreach (var entry in entryManager.EntryList.GetBy(true))
```
Stokalas: Named argument could be used for clarification in line 12

src/DataManagement/RecurringUpdater.cs
```diff
@@ -0,0 +1,125 @@
...
+                return true;
+            }
+
+            ExceptionHandler.Log("Couldn't edit entry in updater:" + editedOriginalEntry);
+            return false;
+        }
+
+
+
+        public static bool AddMonthlyEntry(IEntry entry, int userId, EntryType entryType)
```
Stokalas, line 97: Where do we use `AddMonthlyEntry` since `Update...` seems to be used for already existing entries

Stokalas (approved):
Code seems fine with some improvents.

src/DataManagement/RecurringUpdater.cs
ciliakas: Update Monthly Entry is for private use in class, it used when update a whole Entry List ( basically just the old version of how we used to do updates) and Add Monthly Entry is for standalone use, when we are adding a new entry and it is monthly if that makes sense

src/DataManagement/RecurringUpdater.cs
ciliakas: I tried this, however laterDate - earlierDate returns a time span, which cannot be converted to months, which we are using here

src/DataManagement/RecurringUpdater.cs
Stokalas: So let's say I'm adding an Entry in front code and if it is Monthly expenses, I should call same method as for all expenses or this should be called instead?

ciliakas: At the moment you'd have to manually call this method, but soon I'll integrate it so it you can just add it like normal

Stokalas: Oh ok, then it is awesome

Stokalas (approved):
It will be easier to understand in the future :D Nothing more to say - approving

PR #11, opened 2020-11-05
Ported data calculations and added alternative version

ciliakas:
Ported previous Data Calculations, and temporarilyadded my alternative simplified version of Data Calculations

### DIFF
--- a/src/DataManagement/Saving/AlternativeSavingCalculator.cs
+++ b/src/DataManagement/Saving/AlternativeSavingCalculator.cs
@@ -1,8 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Diagnostics;
-using System.Linq;
-using System.Threading.Tasks;
 using ePiggyWeb.DataManagement.Entries;
 using ePiggyWeb.DataManagement.Goals;
 using ePiggyWeb.Utilities;
--- a/src/DataManagement/Saving/AlternativeSavingCalculator.cs
+++ b/src/DataManagement/Saving/AlternativeSavingCalculator.cs
@@ -1,0 +1,57 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Diagnostics;
+using System.Linq;
+using System.Threading.Tasks;
+using ePiggyWeb.DataManagement.Entries;
+using ePiggyWeb.DataManagement.Goals;
+using ePiggyWeb.Utilities;
+
+namespace ePiggyWeb.DataManagement.Saving
+{
+    public static class AlternativeSavingCalculator
```
Stokalas, line 9: As I understand this does the same and just code improvements?

src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading.Tasks;
```
Stokalas: It could be unused libraries, but what uses `Threading.Tasks`?
`Diagnostics` seems to be unused too.

Stokalas (commented):
Since a lot code is from older project and is already reviewed did not pay much attention. Left few comments on Alternative version, but it seems fine as well.

src/DataManagement/Saving/AlternativeSavingCalculator.cs
ciliakas: Yeah pretty much

src/DataManagement/Saving/AlternativeSavingCalculator.cs
ciliakas: You're correct, they were just unused dependencies, I removed them

Stokalas (approved):
Approving once again.

PR #12, opened 2020-11-05
Authentication

Stokalas:
**Authentication using cookies**
- Login
- Register
- Certain pages only reachable when logged in
- When rerouted to Login after loggin in continues to the same page
-  Different Jumbotron in Index according to whether user is logged in.

### DIFF
--- a/src/Pages/Register.cshtml.cs
+++ b/src/Pages/Register.cshtml.cs
@@ -1,7 +1,13 @@
 using System;
 using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
 using System.Linq;
+using System.Security.Claims;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
+using ePiggyWeb.Authentication;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Authentication.Cookies;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.RazorPages;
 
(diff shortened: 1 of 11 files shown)


src/Pages/Register.cshtml.cs
```diff
@@ -1,16 +1,62 @@
...
+
+        public async Task<IActionResult> OnPost()
         {
+            if (!Password.Equals(PasswordConfirm))
+            {
+                ErrorMessage = "Passwords did not match!";
+                return Page();
+            }
+
+            var Id = UserAuth.Registration(Email, Password);
```
ciliakas: 'Id' should probably start with a lowercase letter

src/Pages/Register.cshtml.cs
```diff
@@ -1,16 +1,62 @@
...
+                 {
+                     new Claim(ClaimTypes.UserData, Id.ToString()),
+                     new Claim(ClaimTypes.Email, Email)
+                 };
+                 var claimsIdentity = new ClaimsIdentity(claims, "Login");
+                 await HttpContext.SignInAsync(CookieAuthenticationDefaults.AuthenticationScheme,
+                     new ClaimsPrincipal(claimsIdentity));
+ 
+                 return Redirect("/Index");
+            };
```
ciliakas: Could you explain how the registration works?  The things that caught my eye were 'async Task' and 'Claim',

src/Pages/logIn.cshtml.cs
```diff
@@ -1,31 +1,49 @@
...
-        public void OnPost()
+
+        public void OnGet()
         {
-            if (UserAuth.Login(Email, Password))
+            ErrorMessage = "";
+        }
+        public async Task<IActionResult> OnPost(string ReturnUrl)
+        {
+            var Id = UserAuth.Login(Email, Password);
```
ciliakas: 'Id' and 'ReturnUrl' should probably start with lowercase

ciliakas (changes requested):
Everything works, the only small glitch that I caught was that if you have your login information saved in browser, if you click register the same info is already filled in into the registration form.
Other small issues in the comments

Stokalas:
**Added fixes:**
- Renamed id and returnUrl variables/parameters
- Turned off autocomplete on Register form
- When trying to access Login/Register when logged in (by link for example: `https://localhost:44330/Login` reroutes to `/Index`

src/Pages/Register.cshtml.cs
Stokalas: - Citing: 
> An async method runs synchronously until it reaches its first await expression, at which point the method is suspended until the awaited task is complete. "

- Claims are data that is saved in cookie and can be accessed as long as person logged in.

- Creating list of claims (data saved in cookie):
` var claims = new List<Claim>
                 {
                     new Claim(ClaimTypes.UserData, Id.ToString()),
                     new Claim(ClaimTypes.Email, Email)
                 };`

- Variable with specified claims, authentication type.
`var claimsIdentity = new ClaimsIdentity(claims, "Login");`

- Creation of cookie (boilerplate code not sure about each step of it):
` await HttpContext.SignInAsync(CookieAuthenticationDefaults.AuthenticationScheme,
                     new ClaimsPrincipal(claimsIdentity));`


- And finally: 

> "IActionResult type. The IActionResult return type is appropriate when multiple ActionResult return types are possible in an action. The ActionResult types represent various HTTP status codes. Any non-abstract class deriving from ActionResult qualifies as a valid return type."

**_Quite a bit to read, but I tried to be as informative as I could._** 🙂

Stokalas: I skipped some parts since I think they are understandable.

ciliakas: Thanks for the in depth answer, clarified this method alot!

ciliakas (approved):
Approving after fixes

PR #13, opened 2020-11-07
Goal display

Stokalas:
**Goal display:**
- Displays goals dinamically by 1,2 or 3 in a row.
- Remove goals
- When savings are >=100% of required amount for goal ability button to remove goal and it to expenses
- Working progress bars
- Modal to enter new goals
- Base to use any user ID, but 0 left as placeholder.

### DIFF
--- a/src/Pages/EditEntry.cshtml.cs
+++ b/src/Pages/EditEntry.cshtml.cs
@@ -44,6 +44,7 @@ namespace ePiggyWeb.Pages
         public void OnPost()
         {
             if (!ModelState.IsValid) return;
+            Entry.Title = Title;
 
             if (EntryTypeInt == 1)
             {
--- a/src/Pages/EditEntry.cshtml.cs
+++ b/src/Pages/EditEntry.cshtml.cs
@@ -37,37 +37,37 @@ namespace ePiggyWeb.Pages
                 Title = Entry.Title;
                 return;
             }
-            Response.Redirect(EntryTypeInt == 1 ? "/Income" : "/Expenses");
+
+            OnPostCancel(); //I entry is empty going back
         }
 
         /*Editing and redirecting according to EntryType*/
-        public void OnPost()
+        public IActionResult OnPost()
         {
-            if (!ModelState.IsValid) return;
+            if (!ModelState.IsValid)
+            {
+                return Page();
+            }
             Entry.Title = Title;
 
             if (EntryTypeInt == 1)
             {
                 EntryDbUpdater.Edit(Entry.Id, Entry, EntryType.Income);
-                Response.Redirect("/Income");
+                return RedirectToPage("/Income");
             }
             else
             {
                 EntryDbUpdater.Edit(Entry.Id, Entry, EntryType.Expense);
-                Response.Redirect("/Expenses");
+                return RedirectToPage("/Expenses");
             }
         }
 
         /*If cancel pressed return to previous page*/
-        public void OnPostCancel()
+        public IActionResult OnPostCancel()
         {
-            if (EntryTypeInt == 1)
-            {
-                Response.Redirect("/Income");
-                return;
-            }
-            Response.Redirect("/Expenses");
+            return RedirectToPage(EntryTypeInt == 1 ? "/Income" : "/Expenses");
         }
+
 
 
     }
--- a/src/Pages/Goals.cshtml
+++ b/src/Pages/Goals.cshtml
@@ -82,7 +82,6 @@ else
                     </button>
                 </div>
                 <div class="modal-body">
-
                     <div asp-validation-summary="ModelOnly" class="text-danger"></div>
                     <input asp-for="Title" type="text" placeholder="Title*" value="" />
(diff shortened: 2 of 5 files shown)


src/Pages/EditEntry.cshtml.cs
```diff
@@ -37,37 +37,38 @@ public void OnGet(int id, int entryType)
...
-        public void OnPostCancel()
+        public IActionResult OnPostCancel()
         {
-            if (EntryTypeInt == 1)
-            {
-                Response.Redirect("/Income");
-                return;
-            }
-            Response.Redirect("/Expenses");
+            return RedirectToPage(EntryTypeInt == 1 ? "/Income" : "/Expenses");
```
ciliakas, line 68: Nice improvement to this whole class, the code is a lot more tidy and easier to understand

src/Pages/Goals.cshtml
```diff
@@ -1,36 +1,108 @@
...
+            {
+                @Html.Raw("</div>")
+            }
+        }
+    </form>
+    if (num % 3 != 0)
+    {
+        @Html.Raw("</div>")
+    }
+}
```
ciliakas: I really like the system you created for displaying of goals, could you explain a bit how it works?

src/Pages/Goals.cshtml.cs
```diff
@@ -1,13 +1,77 @@
...
+
+        [Required(ErrorMessage = "Required")]
+        [BindProperty]
+        public decimal Amount { get; set; }
+
         public void OnGet()
         {
+            var dataManager = new DataManager();
+            Goals = dataManager.Goals.GoalList;
+            UserId = int.Parse(User.FindFirst(ClaimTypes.Name).Value);
```
ciliakas, line 35: Am I understanding correctly that you save and get the user id in the cookies so you don't have to access the database to find it?

src/Pages/Goals.cshtml.cs
```diff
@@ -1,13 +1,77 @@
...
+            DeleteGoalFromDb(id);
+            return RedirectToPage("/goals");
+        }
+
+        public IActionResult OnPostPurchased(int id, string title, string amount)
+        {
+            decimal.TryParse(amount, out var parsedAmount);
+            var entry = new Entry(title, parsedAmount, DateTime.Today, recurring:false, importance:1);
+            EntryDbUpdater.Add(entry, 0, EntryType.Expense);
+            DeleteGoalFromDb(id);
```
ciliakas, line 67: Later on I'll make a method for this so you don't have to implement this logic in the front end itself :)

ciliakas (changes requested):
Good work, I really like the goal display and that we can display as many goals as we want. However a couple of things I think need to be addressed:
- When adding goals, when we leave input fields empty and click add goal, it closes the dialog/form instead of showing what the user inputted wrong ( when you click add goal again, we can see the red error labels)
- This is really minor, but the green color used for the add to expenses button really hurts my eyes, when contrasted with the blue

src/Pages/Goals.cshtml.cs
Stokalas: I wanted to ask you for that, but now just for testing how it works I've written it in front.

src/Pages/Goals.cshtml.cs
Stokalas: Yep, it is saved in these claims, you've asked about in previous pull request.

src/Pages/Goals.cshtml
Stokalas: - Basically it goes through all goals, and when number is divisible by 3 i.e. **0**,3,6... (because max 3 elements in one row) it creates new row:
` @Html.Raw("<div class=\"row p-0 mx-auto mt-3\">")` - just a tag with class that puts items into one line.
- After adding goal, if it's number is divisible by 3 it closes the div, since full row is made (pay attention that closing happens after new item added and incremented, and opening before adding the item:
`if (num % 3 == 0)
            {
                @Html.Raw("</div>")
            }`
- After going through all items if last row was not closed (because it was not filled to have 3 items, it checks that (number is not divisible by 3 and closes last row):
`if (num % 3 != 0)
    {
        @Html.Raw("</div>")
    }`

_This is the logic of displaying all goals "dinamically", if you have questions about other aspects of this code don't hesitate to ask._

Stokalas:
1. I was not able to display Modal on page load after wrong data added. It requires  Javascript etc. And since I'm trying to keep code "C#-arpy" (none JS used by me, except one that comes with bootstrap etc) I did not implement it, just added alert that informs user that something went wrong.
2. Changed color of cart that adds goal to expenses.
3. Progress bar is more colorful than ever: 
   - 0-29% - Red (_Bootstrap: danger_)
   - 30-84% - Yellow  (_Bootstrap: warning_)
   - 85-99% - Electric  (_Bootstrap: info_)
   - 100% - Green (_Bootstrap: success_).

ciliakas (approved):
Colors have been adjusted, and if there's no easy fix for the dialog issue then really it's not worth the effort to spend that much time on a small detail. Good job, approving.

PR #14, opened 2020-11-08
404 error Page

Stokalas:
**404 Error page**

- When wrong route entered rerouting to this page instead of 404 error.
- Navbar sticks at the top when scrolling

_Nothing important, just thought it would be cool. [INSPIRATION](https://www.donaldjtrump.com/asd)_


ciliakas (approved):
Checking on mobile, but the code looks fine, and tested before on pc, the error page looks fine and the error messages are funny :D

PR #15, opened 2020-11-08
User data display

Stokalas:
Displays Income/Expenses, Goals according to authorized user.

### DIFF
--- a/src/Pages/Expenses.cshtml.cs
+++ b/src/Pages/Expenses.cshtml.cs
@@ -47,7 +47,7 @@ namespace ePiggyWeb.Pages
                OnGet();
                return Page();
             }
-
+            UserId = int.Parse(User.FindFirst(ClaimTypes.Name).Value);
             var entry = new Entry(Title, Amount, Date, Recurring, Importance);
             EntryDbUpdater.Add(entry, UserId, EntryType.Expense);
             return RedirectToPage("/Expenses");
@@ -55,6 +55,7 @@ namespace ePiggyWeb.Pages
 
         public IActionResult OnPostDelete(int id)
         {
+            UserId = int.Parse(User.FindFirst(ClaimTypes.Name).Value);
             EntryDbUpdater.Remove(id, UserId, EntryType.Expense);
             return RedirectToPage("/Expenses");
         }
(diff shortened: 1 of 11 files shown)


src/Pages/Expenses.cshtml.cs
```diff
@@ -29,35 +30,34 @@ public class ExpensesModel : PageModel
...
+        public IActionResult OnPostNewEntry()
         {
-            DataManager dataManager;
             if (!ModelState.IsValid)
             {
-                dataManager = new DataManager();
-                Expenses = dataManager.Expenses.EntryList;
-                return;
+               OnGet();
+               return Page();
```
ciliakas, line 48: What does this return?

src/Pages/Goals.cshtml
```diff
@@ -41,7 +41,7 @@ else
                 <h3 class="text-light problematic-title" title="Title">@item.Title </h3>
                 <div class="progress mt-4">
                     @{
-                        var percent = decimal.Round(Model.Savings / item.Amount * 100);
+                        var percent = item.Amount == 0 ? 0 : decimal.Round(Model.Savings / item.Amount * 100);
                         percent = percent > 100 ? 100 : percent;
                         var color = percent < 100 ? (percent < 30? "danger" : (percent < 85 ? "warning" : "info" )) : "success";
```
ciliakas: Space missing after '30' and before '?'.
Also I was thinking, should we declare how many of these ?: statements should we stack in one line? Should two be the limit? Because after we use them too much they become less clear and an switch statement would be clearer

ciliakas (approved):
Everything looks good, except for some minor details in the comments. And nice job on doing the fixes required in the back end to make this all work

src/Pages/Expenses.cshtml.cs
Stokalas: Same page as it was in

src/Pages/Goals.cshtml
Stokalas: Yes, this looked like a good idea, because I wanted to add 3rd color, then when I added 4th it started to get messy, I think I'll use switch here.

PR #17, opened 2020-11-08
Entry update idea

ciliakas:
Changed Entry and Goal creators that used to create objects without id and user id, to a static method CreateLocalEntry/ CreateLocalGoal to hopefully make it more clear what kind of object we want to create. This is just an idea, I'm not sure if this is a better way or not so feedback is very appreciated


src/DataManagement/Entries/Entry.cs
```diff
@@ -17,7 +17,17 @@ public class Entry : IEntry
...
+        public static Entry CreateLocalEntry(string title, decimal amount, DateTime date, bool recurring, int importance)
+        {
+            return new Entry(title, amount, date, recurring, importance);
+        }
+
+        public static Entry CreateLocalEntry(IGoal goal, DateTime date, bool recurring, int importance)
+        {
+            return new Entry(goal.Title, goal.Amount, date, recurring, importance);
+        }
+
```
Stokalas, line 29: This one is nice

Stokalas (commented):
Code seems fine and we can see usage of it in code. But what are the benefits of using this instead of `= new Entry(constructor args)`? Because it seems to do the same thing, just easier to understand by name, but creating new instance of object with arguments to constructor is pretty familiar too.

src/DataManagement/Entries/Entry.cs
ciliakas: Functionally it's exactly the same thing, but I noticed that is pretty confusing sometimes, why we don't need an ID and user ID to call a constructor, and sometimes we do, so this is just purely to make the code less confusing

Stokalas: If we're speaking on this level, I can see why this is an improvement. And as I can see you've already changed code in front, so changes should not cause any problems. Nice.

Stokalas:
Since you used outdated version of Income/Expenses.cshtml there were some merge conflicts, but I think I've solved them 🙂

PR #18, opened 2020-11-09
Dev to Main

Stokalas:
Since there were quite some changes and code I think it is time to merge dev to main, to keep it a bit smaller.

All code should be already removed etc. 
But I think it would be cool if everyone skimmed it once again to see something that maybe has not been seen :) 

Files changed 116, but I think quite a lot of them are files from previous project and auto-generated.

**MERGE THIS ONLY WHEN ALL CURRENT PULL REQUESTS TO DEV ARE MERGED**


ciliakas (approved):
I think it's about time to merge to main, our program has ported a big portion of it's functionality to the web application.

Stokalas (commented):
I think it was kinda silly to ask to review all those files, since there are simply to many of these. I think I will merge this for now.

PR #19, opened 2020-11-09
Goal Database Update

ciliakas:
Part of back end restructuring.
The thing to note is that this is written in mind that the front end can use both DataManager and GoalDatabase. Method calls should stay more or less the same as before.
Changes:
- Renamed GoalDbUpdater to GoalDatabase (hence why you can't see that changes in the class)
Features:
Goal manager:
- Remove List of Goals, by id array, by goal list
- Remove goal by id
- Add range
- Edit by id
- Move goal to expenses
Goal database:
- All of the above
- Read: Goal, Goal list, by id or by predicate
- Delete by predicate


src/DataBase/GoalDatabase.cs
```diff
@@ -0,0 +1,162 @@
...
+        {
+            return Delete(goal.Id, goal.UserId);
+        }
+
+        public static bool Delete(int id, int userId)
+        {
+            return Delete(x => x.Id == id && x.UserId == userId);
+        }
+
+        public static bool Delete(Func<GoalModel, bool> filter)
```
Stokalas, line 72: How is this stuff called? Delegates or smth? Or is this just simple passing function as parameter?

src/DataBase/GoalDatabase.cs
```diff
@@ -0,0 +1,162 @@
...
+            {
+                ExceptionHandler.Log(ex.ToString());
+                return false;
+            }
+            return true;
+        }
+
+        public static bool DeleteList(IEnumerable<IGoal> goalList)
+        {
+            var enumerable = goalList as IGoal[] ?? goalList.ToArray();
```
Stokalas, line 91: Can you explain this line?

src/DataBase/GoalDatabase.cs
```diff
@@ -0,0 +1,162 @@
...
+                db.SaveChanges();
+                return dbExpense.Id;
+            }
+            catch (InvalidOperationException ex)
+            {
+                ExceptionHandler.Log(ex.ToString());
+                ExceptionHandler.Log("Couldn't find goal id: " + goalId + " in database");
+                return -1;
+            }
+        }
```
Stokalas, line 142: This one I needed in front very much, nice!

src/DataManagement/Goals/GoalManager.cs
```diff
@@ -19,59 +22,116 @@ public GoalManager(IGoalList goalList, int userId = 0)
...
         {
-            var id = GoalDbUpdater.Add(goal, UserId);
+            var id = GoalDatabase.Create(goal, UserId);
             //Check if id is correct, return false if something is wrong
-
             goal.Id = id;
-            goal.UserId = UserId;
             GoalList.Add(goal);
-            return false;
+            return true;
```
Stokalas, line 35: This change is interesting, why false removed and changed to true?

src/DataManagement/Goals/GoalManager.cs
```diff
@@ -19,59 +22,116 @@ public GoalManager(IGoalList goalList, int userId = 0)
...
+            if (!GoalDatabase.Update(id, UserId, updatedGoal))
             {
                 return false;
             }
-            var localGoal = GoalList.FirstOrDefault(x => x.Id == oldGoal.Id);
+            var localGoal = GoalList.FirstOrDefault(x => x.Id == id);
             if (localGoal is null)
             {
-                ExceptionHandler.Log("Edited goal id: " + oldGoal.Id + " in database but couldn't find it locally");
+                ExceptionHandler.Log("Edited goal id: " + id + " in database but couldn't find it locally");
```
Stokalas, line 65: Just a thought: this code is checked with if and will not raise exception of any sort, so it is kinda strange to use ExceptionHandler when there is no really no exception (it is controlled with control flow).

src/DataManagement/Goals/GoalManager.cs
```diff
@@ -19,59 +22,116 @@ public GoalManager(IGoalList goalList, int userId = 0)
...
+        {
+            var idArray = idList as int[] ?? idList.ToArray();
+
+            if (!GoalDatabase.DeleteList(idArray, UserId))
+            {
+                return false;
+            }
+
+            var temp = new GoalList();
+            temp.AddRange(from entry in GoalList from id in idArray where entry.Id == id select entry);
```
Stokalas: why is this used as querie and not simple linq?

Stokalas (approved):
Haven't tested but the code seems fine. Just some parts I did not understand, so I asked in code comments.

src/DataBase/GoalDatabase.cs
ciliakas: Yeah, it's a delegate of Func type, takes in one parameter (GoalModel) and returns a result (bool). From line calling it with a lambda expression, but you could write in a different form, as long as it matches the Func<GoalModel,bool> type.
This link explains it more in depth: https://docs.microsoft.com/en-us/dotnet/api/system.func-2?view=netcore-3.1
In general I'm using this to pass a predicate to a method

src/DataManagement/Goals/GoalManager.cs
ciliakas: This was a mistake from before.
In general the Data Managers and Database classes lack proper exception handling. So as you see, some of them are returning true always, but I want to make them return false when something went wrong

src/DataManagement/Goals/GoalManager.cs
ciliakas: So this code stands for :
            foreach (var goal in GoalList)
            {
                foreach (var id in idArray)
                {
                    if (goal.Id == id)
                    {
                        temp.Add(goal);
                    }
                }
            }
It's kind of like joining two tables together, so I don't know how you'd have to write it as a usual lambda expression we use (this is code is reshaper by the way :D ). If you know how to write this in a simpler way I'm open to suggestions, as I'm really no expert in this subject

src/DataManagement/Goals/GoalManager.cs
ciliakas: Good point actually. What I wanted to do was to know if something went wrong at this place in code, and the exception handler seemed like the way to do it? Maybe I should actually throw an exception here if something goes wrong, and surround it with a try catch block so we can actually print stack trace and all that?

src/DataManagement/Goals/GoalManager.cs
Stokalas: Something like LINQ join which was covered in yesterday's lecture may be the solution?

src/DataBase/GoalDatabase.cs
ciliakas: First I'm trying to cast the passed goalList parameter as an array of IGoals, but if it can't be cast then instead an array of IGoals is created (not cast). It's a bit confusing, but this was resharpers suggestion and it does make sense if you think about it (thanks for the question, actually investigated it and now I understand better myself :D)

src/DataManagement/Goals/GoalManager.cs
ciliakas: Actually it turns out this is (as I understand) the is the LINQ stuff that was covered in yesterday's lecture
Our code:

            temp.AddRange(
                from goal in GoalList 
                from id in idArray 
                where goal.Id == id 
                select goal);

Example given on LINQ Join:

            var personCountryInner = from p in persons
                join c in regions
                    on p.RegionId equals c.ID
                select new
                {
                    PersonName = p.Name,
                    RegionName = c.Name
                };

Stokalas (approved):
Seems fine.

PR #20, opened 2020-11-09
Entry Database Update

ciliakas:
Final part of back end refactoring.
Both EntryDatabase and DataManager work and can be used for the front end, DataManager is designed more in mind with working in an app environment and not the web. Front end has been updated to match the backed, I did minor testing and it seems to work, but more testing wouldn't hurt.
In general use of back end shouldn't have changed much, except for that I renamed EntryDatabase methods were renamed to match CRUD ( create, read, update, delete)
Features:
-Read Entry, EntryList by id and predicate
-Delete by id and predicate
-Add monthly enter now automatically adds recurring entries if needed
-Edit monthly now adds recurring entries if needed


src/Pages/Income.cshtml
```diff
@@ -93,7 +93,7 @@
                         </td>
                         <td class="border-primary">
                             <a asp-page="/EditEntry" asp-route-id="@item.Id" asp-route-entryType="1" class="btn btn-sm btn-success">Edit </a>
-                            <input asp-page-handler="Delete" asp-route-id="@item.Id" type="submit" class="btn btn-sm btn-danger" value="Delete" />
+                            <input asp-page-handler="Delete" asp-route-id="@item.Id" asp-route-userId="@item.UserId"  type="submit" class="btn btn-sm btn-danger" value="Delete" />
```
Stokalas, line 96: Why was this added? Same in expenses.cshtml

src/DataManagement/Entries/EntryManager.cs
```diff
@@ -84,42 +92,43 @@ public bool Edit(int id, IEntry newEntry)
...
+
             EntryList.Remove(temp);
             return true;
         }
 
         public bool RemoveAll(IEntryList entryList)
         {
-            var idList = entryList.Select(va => va.Id).ToArray();
-            if (!EntryDbUpdater.RemoveAll(idList, entryList.EntryType))
+            var idArray = entryList.Select(va => va.Id).ToArray();
```
Stokalas: Random question, why "va" is used? :D

src/DataManagement/RecurringUpdater.cs
```diff
@@ -12,114 +11,38 @@ public static bool UpdateRecurring(IEntryManager entryManager)
...
-                }
+                if (differenceInMonths <= 0) continue;
+                entry.Date = entry.Date.AddMonths(differenceInMonths);
+                entryManager.Edit(entry.Id, entry);
             }
             return true;
         }
 
-        private static bool UpdateEntry(IEntryManager entryManager, IEntry entry, int months)
+        public static IEntryList CreateRecurringListWithoutOriginalEntry(IEntry entry, EntryType entryType)
```
Stokalas, line 21: So this now creates list with all needed entries and adds them then, if I understood corectly?

src/Pages/Expenses.cshtml.cs
```diff
@@ -49,14 +49,14 @@ public IActionResult OnPostNewEntry()
             }
             UserId = int.Parse(User.FindFirst(ClaimTypes.Name).Value);
             var entry = Entry.CreateLocalEntry(Title, Amount, Date, Recurring, Importance);
-            EntryDbUpdater.Add(entry, UserId, EntryType.Expense);
+            EntryDatabase.Create(entry, UserId, EntryType.Expense);
```
Stokalas, line 52: Thanks for renaming it for me :D Im just a bit concerned that because I changed these parts a bit in my new branch there will be conflict, but nonetheless its just one line so everything should be fine :)

Stokalas (approved):
Code seems fine, I've left few comments. Only problem I see is that when code builds up in front and elsewhere these changes require to change quite a bit of code, but if as I understand it's coming to the end it is cool 🙂

src/Pages/Income.cshtml
ciliakas: Beforehand we had I believe goals requiring both id and user id to delete, so I made it standard that to delete something from the database you need to pass both the user id and the id.
This was added to the front end to pass this the user id as a variable to the Delete method, but maybe there's another way to do this without adding code to the front end?

src/DataManagement/RecurringUpdater.cs
ciliakas: The methods in recurring updater are now utilities only, the UpdateRecurring method updates the entries inside an Entry Manager, but the other two methods just form lists of entries that need to be added when a recurring entry is added or editied. This class won't need to be accessed by the front end

src/Pages/Expenses.cshtml.cs
ciliakas: Last change, I promise :D Sorry about these names, I just can't seem to make my mind on the best name for them, but I will just stick to what we have in here from here on out to stop the endless change

src/DataManagement/Entries/EntryManager.cs
ciliakas: Have no idea how that's in here, will add a fix for these :D

Stokalas (approved):

PR #21, opened 2020-11-09
Income/expenses filter + graphs

Stokalas:
Creating pull request for testing purposes, not everything is developed.
 - Ability to filter Income/expenses by Date period
 - On load displays only current month data
 - Only parts where it will lose the filter is when entry is added/removed. (a bit hard to solve since everything uses forms and they reset each other)


ciliakas:
Maybe you could save the state of the filter in the cookies?

Stokalas:
It is quite interesting idea, but I'm not sure how "ethical" it is to store data inside cookies, that may change a dozen of times during one use of app.

Stokalas:
Since I only wanted to play around with, but it worked fine, so eventough it is bad practice to include, but graphs are also inside this branch :)

ciliakas:
Is this pull request ready for review? :)

Stokalas:
Yup, It is waiting to be reviewed and changed if needed :)

src/Pages/IncomeGraph.cshtml
```diff
@@ -0,0 +1,55 @@
...
+            'titleTextStyle': { color: '#FFF' },
+            'hAxis': { 'textStyle':{'color': '#FFF'} }
+        };
+
+        var chartPie = new google.visualization.PieChart(document.getElementById('graphPie'));
+        var chartColumn = new google.visualization.ColumnChart(document.getElementById('graphColumn'));
+        chartPie.draw(data, options);
+        chartColumn.draw(data, options);
+    }
+</script>
```
DeividasLT5, line 70: Very noticable red symbol says "No newline at end of file" made me wonder if we have and should we have any standard for this or anything?

src/Pages/ComparisonGraph.cshtml
```diff
@@ -0,0 +1,50 @@
...
+    // Load google charts
+    google.charts.load('current', { 'packages': ['corechart'] });
+    google.charts.setOnLoadCallback(drawChart);
+
+    // Draw the chart and set the chart values
+    function drawChart() {
+        var data = google.visualization.arrayToDataTable([
+            ['Type', 'Amount'],
+            ['Expenses', @Model.Expenses.ToString(CultureInfo.InvariantCulture)],
+            ['Income',  @Model.Income.ToString(CultureInfo.InvariantCulture)]
```
ciliakas, line 41: Could you explain why you're using invariant culture here?

src/Pages/ComparisonGraph.cshtml.cs
```diff
@@ -0,0 +1,44 @@
...
+        private int UserId { get; set; }
+        [BindProperty]
+        public DateTime StartDate { get; set; }
+        [BindProperty]
+        public DateTime EndDate { get; set; }
+        public void OnGet()
+        {
+            var today = DateTime.Now;
+            StartDate = new DateTime(today.Year, today.Month, 1);
+            EndDate = DateTime.Today;
```
ciliakas, line 26: Maybe it would be better to see entries until the end of the month? I guess this is subjective though

src/Pages/ExpensesGraph.cshtml
```diff
@@ -0,0 +1,63 @@
...
+
+    // Draw the chart and set the chart values
+    function drawChart() {
+        var data = google.visualization.arrayToDataTable([
+            ['Importance', 'Amount'],
+            ['Necessary', @Model.Expenses.GetBy(Importance.Necessary).GetSum().ToString(CultureInfo.InvariantCulture)],
+            ['High', @Model.Expenses.GetBy(Importance.High).GetSum().ToString(CultureInfo.InvariantCulture)],
+            ['Medium', @Model.Expenses.GetBy(Importance.Medium).GetSum().ToString(CultureInfo.InvariantCulture)],
+            ['Low', @Model.Expenses.GetBy(Importance.Low).GetSum().ToString(CultureInfo.InvariantCulture)],
+            ['Unnecessary', @Model.Expenses.GetBy(Importance.Unnecessary).GetSum().ToString(CultureInfo.InvariantCulture)]
```
ciliakas, line 53: Leaving a comment here even though I commented up above, we have a NumberFormatter utility function to display currency for every locale, if it's used here to display currency.

ciliakas: Or you could you use this:
https://docs.microsoft.com/en-us/globalization/locale/number-formatting-in-dotnet-framework
Which I guess would make more sense, I'm not sure if the number formatting I've created before is actually useful :D

ciliakas (changes requested):
The code looks neat, I've left a few questions in the comments but they are not really that pressing.
Overall the graphs looks nice and the filtering works great, this is some good progress.
There are a few errors I did notice though:
- I can select the start date to be later then the end date
- Maybe if the graphs are empty, they should show some placeholder info or some text, so the user would know that everything worked fine, there just was no entries during his selected time span
- I think 'saved this period' in Incomes should be renamed to 'earned' because it seems easy to confuse this with balance instead of total income
- If we get an error when adding an entry, all fields reset, including the filter ones, which then show data for year 0 :D 
Screenshots for clarity:
back to the future
dingo entries
should prolly be earned

src/Pages/ComparisonGraph.cshtml.cs
Stokalas: I just took _Swedbank_ filtering by date system as an example, and they don't even allow to go to the future :D, but I just wanted to set from 1st of Month till today period as preset value. And if user wants to see data until the end of the month, they can set that in filter.

src/Pages/ComparisonGraph.cshtml
Stokalas: when I use plain decimal or smth it gives me number with comma and the graph won't render it, so only solution I found was to convert to string with invariant culture.

src/Pages/ExpensesGraph.cshtml
Stokalas: Will take a look at it, thanks!

Stokalas: Since this is for graph numbers and not only displaying, this should not work as I understand?

src/Pages/IncomeGraph.cshtml
Stokalas: Could you specify where is the red symbol, because for me it doesn't show anything.

src/Pages/ComparisonGraph.cshtml.cs
ciliakas: Yeah, that makes sense

src/Pages/ExpensesGraph.cshtml
ciliakas: So do these graphs take in numbers as strings? I'm a bit confused :D

Stokalas: It doesn't matter for this JS code if its just decimal with ".", or decimal inside qoutes, what matters is that it has to have dot :D

ciliakas: Oh, so I guess that does mean it can't be displayed as currency. Then the way you've set it up here is really the only way to do it :D

ciliakas (approved):
All issues seems to be addressed, good work

src/Pages/ExpensesGraph.cshtml
Stokalas: Well I don't like this part myself, but it works and doesn't seem to increase loading time very much so I left it.

src/Pages/IncomeGraph.cshtml
DeividasLT5: It was just after the text on line 55, though it doesn't show it here.

DeividasLT5: It was just after the text on line 55, though it doesn't show it here.

DeividasLT5 (approved):
Seems like everything is resolved, I'm happy with it

PR #22, opened 2020-11-11
"Password-ing"

Stokalas:
- Ability to recover forgotten password
- Ability to change password when logged in
- When password recovery code sent but password not changed reminder in index.
- Password recovery code is valid: 15min; until logged in/registered; until recovery canceled;
- Password recovery page should not be reachable when recovery is not in progress.
- Change password page should not be reachable when not logged in.

### DIFF
--- a/src/Pages/ForgotPassword.cshtml.cs
+++ b/src/Pages/ForgotPassword.cshtml.cs
@@ -30,6 +30,8 @@ namespace ePiggyWeb.Pages
 
         public string ErrorMessage = "";
 
+        public bool CodeSent = false;
+
 
         public IActionResult OnGet()
         {
@@ -37,11 +39,18 @@ namespace ePiggyWeb.Pages
             {
                 return RedirectToPage("/index");
             }
+
+            if (Request.Cookies.ContainsKey("recoveryCode"))
+            {
+                return Page();
+            }
             Email = Request.Cookies["Email"];
             var recoveryCode = EmailSender.SendRecoveryCode(Email).ToString();
 
             var option = new CookieOptions() { Expires = DateTime.Now.AddMinutes(15) };
             Response.Cookies.Append("recoveryCode", recoveryCode, option);
+            CodeSent = true;
+
             return Page();
         }
 
@@ -72,6 +81,13 @@ namespace ePiggyWeb.Pages
             }
 
             return Page();
+        }
+
+        public IActionResult OnPostCancel()
+        {
+            Response.Cookies.Delete("recoveryCode");
+            Response.Cookies.Delete("Email");
+            return RedirectToPage("/login");
         }
 
     }
--- a/src/Pages/ForgotPassword.cshtml.cs
+++ b/src/Pages/ForgotPassword.cshtml.cs
@@ -1,0 +1,76 @@
+using System;
+using System.ComponentModel.DataAnnotations;
+using System.Diagnostics;
+using System.Security.Claims;
+using ePiggyWeb.DataBase;
+using ePiggyWeb.Utilities;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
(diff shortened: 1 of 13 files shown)


Stokalas:
**Note to myself**: to add client side input validation before merging this pull request 🙂

src/Pages/logIn.cshtml.cs
```diff
@@ -56,8 +56,8 @@ public async Task<IActionResult> OnPost(string returnUrl)
         public IActionResult OnPostForgotPassword()
         {
             if (!ModelState.IsValid) return Page();
-
-            Response.Cookies.Append("Email", Email);
+            var option = new CookieOptions() { Expires = DateTime.Now.AddMinutes(15) };
```
DeividasLT5, line 82: Just wondering, how does this expiration work? In a little more detail, please

src/Pages/Index.cshtml
```diff
@@ -4,7 +4,16 @@
...
+@if(Request.Cookies.ContainsKey("recoveryCode"))
+{
+    <div class="alert alert-warning" role="alert">
+        Password recovery is in progress! <a asp-page="/forgotPassword" class="alert-link">Go to recovery</a> to finish it!.
+        <button type="button" class="close" data-dismiss="alert" aria-label="Close">
+            <span aria-hidden="true">&times;</span>
+        </button>
+    </div>
+   
+}
```
ciliakas, line 16: Not sure if this is supposed to work this way, but if you start password recovery, and then login without finishing it, this popup doesn't appear

src/Pages/logIn.cshtml.cs
```diff
@@ -43,14 +52,25 @@ public async Task<IActionResult> OnPost(string returnUrl)
                 await HttpContext.SignInAsync(CookieAuthenticationDefaults.AuthenticationScheme,
                     new ClaimsPrincipal(claimsIdentity));
 
+                Response.Cookies.Delete("recoveryCode");
+                Response.Cookies.Delete("Email");
                 return Redirect(returnUrl ?? "/Index");
-                
-
             };
```
ciliakas: Super minor thing, but this semicolon is unnecessary :D

src/Pages/ForgotPassword.cshtml.cs
```diff
@@ -0,0 +1,94 @@
...
+        [BindProperty]
+        [DataType(DataType.Password)]
+        public string PasswordConfirm { get; set; }
+
+        [BindProperty]
+        public string Email { get; set; }
+
+        public string ErrorMessage = "";
+
+        public bool CodeSent = false;
```
ciliakas: Resharper is telling me it's unnecessary to set CodeSent default value.

ciliakas (changes requested):
Neato, good feature to have. Code looks good, left a few minor questions in the comments.
Few errors I've managed to spot though:
- The red error labels for the change password form are hard to read and kind of hurt the eyes
- If I start the password recovery, then login in another tab, and then finish password recovery, I get a blank screen and the password doesn't change. Granted, this is really unlikely to happen, but an error message here or something would be good to have.
Screenshots:
got this by calling recovery then changing then finishing recovery
text is hard to read

src/Pages/Index.cshtml
Stokalas: Yup, it shouldn't since it resets recover if logged in.

src/Pages/ForgotPassword.cshtml.cs
Stokalas: I guess bools by default are assigned to false.

src/Pages/logIn.cshtml.cs
Stokalas: It just an option you can add, how the cookies are handled in the deep back I don't really know :D
I guess it just has a date, from which it is deleted.

ciliakas (approved):
All issues fixed

PR #23, opened 2020-11-13
Email sender update

ciliakas:
Updates to email sender:
- Made the class non static ( suggestion of the lecturer)
- Made email sender async ( otherwise it wouldn't work)
- Added dependency injection for email sender
- Added greeting and farewell emails

### DIFF
--- a/src/Pages/ForgotPassword.cshtml.cs
+++ b/src/Pages/ForgotPassword.cshtml.cs
@@ -1,5 +1,6 @@
 using System;
 using System.ComponentModel.DataAnnotations;
+using System.Threading.Tasks;
 using ePiggyWeb.DataBase;
 using ePiggyWeb.Utilities;
 using Microsoft.AspNetCore.Http;
@@ -34,7 +35,7 @@ namespace ePiggyWeb.Pages
         public bool Expired;
 
 
-        public IActionResult OnGet()
+        public async Task<IActionResult> OnGet()
         {
             if (!Request.Cookies.ContainsKey("Email") || User.Identity.IsAuthenticated)
             {
@@ -46,7 +47,8 @@ namespace ePiggyWeb.Pages
                 return Page();
             }
             Email = Request.Cookies["Email"];
-            var recoveryCode = EmailSender.SendRecoveryCode(Email).ToString();
+            var emailSender = new EmailSender();
+            var recoveryCode = (await emailSender.SendRecoveryCodeAsync(Email)).ToString();
 
             var option = new CookieOptions() { Expires = DateTime.Now.AddMinutes(15) };
             Response.Cookies.Append("recoveryCode", recoveryCode, option);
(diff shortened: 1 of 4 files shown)


src/Pages/ForgotPassword.cshtml.cs
```diff
@@ -46,7 +52,7 @@ public IActionResult OnGet()
                 return Page();
             }
             Email = Request.Cookies["Email"];
-            var recoveryCode = EmailSender.SendRecoveryCode(Email).ToString();
+            var recoveryCode = (await EmailSender.SendRecoveryCodeAsync(Email)).ToString();
```
Stokalas, line 55: This seems nice improvement since it freezed when loading this page for a bit, did this fix it, if you have tested?

src/Startup.cs
```diff
@@ -31,6 +32,7 @@ public void ConfigureServices(IServiceCollection services)
                     options.SlidingExpiration = true;
 
                 });
+            services.AddScoped<EmailSender>();
```
Stokalas, line 35: This is some kind of DI as I understand, but for example in `Register.cshtml.cs` instance of object is created. So shortly how does it work and what does it give? (source to read about it myself is welcome as well).

src/Utilities/EmailSender.cs
```diff
@@ -1,50 +1,69 @@
...
                 EnableSsl = true,
-                Port = PortConst,
                 DeliveryMethod = SmtpDeliveryMethod.Network,
-                Credentials = new NetworkCredential(EmailAddress, EmailPassword)
+                Credentials = new NetworkCredential(OurEmail, EmailPassword)
             };
-            smtp.Send(message);
-
-            return randomCode;
+            await smtp.SendMailAsync(message);
```
Stokalas, line 66: Just curious, did these await everywhere improve the speed? I understand that it protects sender from garbage collection, but what about performance?

Stokalas (commented):
Code seems fine, left a few questions.

src/Pages/ForgotPassword.cshtml.cs
ciliakas: I tested it out a bunch of times, and I can't say for sure if the performance has improved noticeably. However as I understand this is a must if multiple users would be using the website at once, because they couldn't all access the same static class at the same time, because smtp client is not thread safe. Could be wrong here though

src/Startup.cs
ciliakas: Yeah, this is DI. I'll try to explain it in a few words, but I'm not sure I will be able because it's a bit complex. Basically instead of using constructors everywhere in the model, on it being created the needed object are already passed to the model. We can also add the different lifetime of the objects, and so on. As I understand this is used to make the classes less tightly coupled, as well as more efficient usage of Database contexts for example

Sources I've relied on mostly:
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-5.0
https://www.youtube.com/watch?v=HXKGKvoQ0Xs&t=329s
https://stackoverflow.com/questions/35367127/asp-net-mvc-guidelines-for-static-classes-for-database-access

src/Utilities/EmailSender.cs
ciliakas: I don't think it has impacted performance very much.

PR #24, opened 2020-11-13
User database update

ciliakas:
According to the suggestions of the lecturer, started the conversion from static database methods to instance based ones.
Changes:
- Added a new database context class PiggyDbContext for the new instance based classes, but left the old DatabaseContext as well to keep everything else working as to not make as a big a change in a single pull request
- Now the methods associated with UserDatabase are all async
Features:
- Added dependency injection for PiggyDbContext and UserDatabase, so they can easily be used by the front end
- Added delete methods to UserDatabase


src/DataBase/PiggyDbContext.cs
```diff
@@ -0,0 +1,19 @@
...
+    public class PiggyDbContext : DbContext
+    {
+        public DbSet<UserModel> Users { get; set; }
+        public DbSet<ExpenseModel> Expenses { get; set; }
+        public DbSet<IncomeModel> Incomes { get; set; }
+        public DbSet<GoalModel> Goals { get; set; }
+        public PiggyDbContext(DbContextOptions<PiggyDbContext> options) : base(options) {}
+
+    }
+}
```
Stokalas, line 19: Was this file just renamed, or new one created?

src/DataBase/UserDatabase.cs
```diff
@@ -43,36 +51,60 @@ public static int Authenticate(string email, string pass)
...
+        {
+            var temp = await Database.Users.FirstOrDefaultAsync(filter);
+            if (temp is null)
+            {
+                return false;
+            }
+            Database.Users.Remove(temp);
+            await Database.SaveChangesAsync();
             return true;
         }
```
Stokalas, line 93: Where do you see this to be used?

src/DataBase/UserDatabase.cs
```diff
@@ -43,36 +51,60 @@ public static int Authenticate(string email, string pass)
...
 
-        public static UserModel GetUser(int userId)
+        public async Task<UserModel> GetUserAsync(int userId)
+        {
+            return await GetUserAsync(x => x.Id == userId);
+        }
+
```
Stokalas, line 99: What piece of info this returns?

src/appsettings.json
```diff
@@ -6,5 +6,8 @@
       "Microsoft.Hosting.Lifetime": "Information"
     }
   },
+  "ConnectionStrings": {
+    "DefaultConnection": "Server=51.75.187.147;Database=SmartSaver;User Id=usern;Password=123456789;"
+  },
```
Stokalas, line 11: Nice change indeed, but I think this will be moved to user config file 🙂

src/Startup.cs
```diff
@@ -31,6 +33,14 @@ public void ConfigureServices(IServiceCollection services)
                     options.SlidingExpiration = true;
 
                 });
+
+            services.AddDbContext<PiggyDbContext>(options =>
+                options.UseSqlServer(Configuration.GetConnectionString("DefaultConnection")));
+            services.AddScoped(provider =>
+            {
+                var db = provider.GetService<PiggyDbContext>();
+                return new UserDatabase(db);
```
Stokalas, line 43: How does this work? Especially the return is interesting?

Stokalas (commented):
Left a few questions, code seems to be fine.

src/DataBase/PiggyDbContext.cs
ciliakas: This is a new one, I've left the old Database Context intact for now.

ciliakas: Granted it's almost the same as old Database context, with minor differences in it's creation

src/DataBase/UserDatabase.cs
ciliakas: Remember when I was thinking where to use Events? One of the places I figured would be on user deletion, we could send out events to delete all info related to the user. Even if we don't implement that idea, this didn't really take too much time to write :D

src/DataBase/UserDatabase.cs
ciliakas: This returns UserModel ( I wanted to make this as reusable as possible, for example instead of having two different methods to get user id or user email, you just get this user model object and you can extract all the info related from the user from it)

src/Startup.cs
ciliakas: So I've got this from a tutorial on Youtube, so bear that in mind, but as I understand, if we want to inject an object which does not have an empty constructor ( like UserDatabse, which requires a database context), first we need to get the parameters we will be adding to the constructor, so we take the required parameter/service ( in this place PiggyDbContext) and add it into the constructor. I don't know if I explained this very clearly :D

src/DataBase/UserDatabase.cs
Stokalas: I understood user deletion it is usable, but I meant this exact method that takes anonymous function as parameter.

ciliakas: Oh, sorry. Just figured there might be a situation when we have some specific predicates we might be removing by, so it's here just in case.

PR #25, opened 2020-11-14
Entry database update

ciliakas:
Refactored Entry Database to be instance based instead of static and changed all of front end to use Entry Database instead of Data Manager ( for now at least, needed to test how everything works, if I decide to refactor data manager as well, then I'll reintegrate Data Manager)
Changes are pretty similar to the previous User database update, the old Entry Database class was renamed as EntryDatabaseOld (it's still in use by Data Manager).


src/DataBase/EntryDatabase.cs
```diff
@@ -139,69 +140,86 @@ public static bool Delete(Func<IEntryModel, bool> filter, EntryType entryType)
...
             {
-                var entriesToRemove = idArray.Select(id => db.Incomes.FirstOrDefault(x => x.Id == id && x.UserId == userId)).ToList();
-                db.Incomes.RemoveRange(entriesToRemove);
+                var list = new List<IncomeModel>();
+                foreach (var id in idArray)
+                {
+                    var temp = await Database.Incomes.FirstOrDefaultAsync(x => x.Id == id && x.UserId == userId);
+                    list.Add(temp);
+                }
+                Database.Incomes.RemoveRange(list);
```
Stokalas: Why was this changed from LINQ to foreach?

src/DataBase/EntryDatabase.cs
```diff
@@ -1,21 +1,25 @@
...
 using ePiggyWeb.DataManagement.Entries;
 using ePiggyWeb.Utilities;
+using Microsoft.EntityFrameworkCore;
 
 namespace ePiggyWeb.DataBase
 {
-    public static class EntryDatabase
+    public class EntryDatabase
     {
-        /*
```
Stokalas, line 13: So in short this file was changed from static to use async/await?

src/DataBase/EntryDatabaseOld.cs
```diff
@@ -0,0 +1,207 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using ePiggyWeb.DataBase.Models;
+using ePiggyWeb.DataManagement;
+using ePiggyWeb.DataManagement.Entries;
+using ePiggyWeb.Utilities;
+
+namespace ePiggyWeb.DataBase
+{
+    public static class EntryDatabaseOld
```
Stokalas, line 11: Won't review this since, as I understand it is just to save old code.

src/Pages/EditEntry.cshtml.cs
```diff
@@ -22,16 +23,20 @@ public class EditEntryModel : PageModel
...
         {
             var userId = int.Parse(User.FindFirst(ClaimTypes.Name).Value);
-            var dataManager = new DataManager(userId);
             EntryTypeInt = entryType;
 
             Entry = entryType == 1
-                ? (Entry)dataManager.Income.EntryList.FirstOrDefault(x => x.Id == id)
-                : (Entry)dataManager.Expenses.EntryList.FirstOrDefault(x => x.Id == id);
+                ? (Entry) await EntryDatabase.ReadAsync(id, userId, EntryType.Income)
+                : (Entry) await EntryDatabase.ReadAsync(id, userId, EntryType.Expense);
```
Stokalas, line 39: Seems much cleaner now, eventhough it's testing file :D

src/Pages/ForgotPassword.cshtml.cs
```diff
@@ -36,10 +36,10 @@ public class ForgotPasswordModel : PageModel
 
         private EmailSender EmailSender { get; }
         private UserDatabase UserDatabase { get; }
-        public ForgotPasswordModel(EmailSender emailSender, UserDatabase userDatabase)
+        public ForgotPasswordModel(UserDatabase userDatabase, EmailSender emailSender)
```
Stokalas: Just interested, why was this changed?

src/Startup.cs
```diff
@@ -42,6 +42,11 @@ public void ConfigureServices(IServiceCollection services)
                 var db = provider.GetService<PiggyDbContext>();
                 return new UserDatabase(db);
             });
+            services.AddScoped(provider =>
+            {
+                var db = provider.GetService<PiggyDbContext>();
+                return new EntryDatabase(db);
+            });
             services.AddScoped<EmailSender>();
```
Stokalas: As always, nice to see DI happening

Stokalas (approved):
Seems to be well done, left few questions, as always.

src/DataBase/EntryDatabase.cs
ciliakas: Yeah, in essence

src/DataBase/EntryDatabase.cs
ciliakas: Wanted to write this as quick possible, and wasn't sure how to change this expression to be async. Spent some time and figured out how to write as a linq expression, so this was a good question to ask :D

src/Pages/ForgotPassword.cshtml.cs
ciliakas: No reason really I think, just being ocd about pointless things :D

src/DataBase/EntryDatabase.cs
```diff
@@ -142,49 +142,30 @@ public async Task<bool> DeleteAsync(Expression<Func<IEntryModel, bool>> filter,
...
-            else
-            {
-                var entriesToRemove = await Database.Expenses.Where(filter).Cast<ExpenseModel>().ToListAsync();
-                Database.Expenses.RemoveRange(entriesToRemove);
-            }
-
+                EntryType.Income => await Database.Incomes.Where(filter).Cast<IncomeModel>().ToListAsync(),
+                _ => await Database.Expenses.Where(filter).Cast<ExpenseModel>().ToListAsync()
+            };
+            Database.RemoveRange(entriesToRemove);
```
Stokalas: How does this work, or source from where to read about?

src/DataBase/EntryDatabase.cs
```diff
@@ -142,49 +142,30 @@ public async Task<bool> DeleteAsync(Expression<Func<IEntryModel, bool>> filter,
...
-                }
-                Database.Expenses.RemoveRange(list);
-            }
-
+                EntryType.Income => await Task.WhenAll(idArray.Select(selector: async id =>
+                    await Database.Incomes.FirstOrDefaultAsync(x => x.Id == id && x.UserId == userId))),
+                _ => await Task.WhenAll(idArray.Select(selector: async id =>
+                    await Database.Expenses.FirstOrDefaultAsync(x => x.Id == id && x.UserId == userId)))
+            };
+            Database.RemoveRange(entriesToRemove);
```
Stokalas: Same here

Stokalas (approved):
Nice to see, that you managed to keep Linq with async, just a few questions. Approving

src/DataBase/EntryDatabase.cs
ciliakas: I think line is more clear when I removed the (as I understand) redundant cast

     await Database.Incomes.Where(filter).ToListAsync()


Source to read up on: https://docs.microsoft.com/en-us/dotnet/api/system.data.entity.queryableextensions.tolistasync?view=entity-framework-6.2.0

src/DataBase/EntryDatabase.cs
ciliakas: Thank you for asking about this, because this led me to some big improvements for this part :D Before this was generating threading issues, I think it's because of the two await statements, but I'm not completely sure why. 
Now the code looks like this:

            var filter = PredicateBuilder.BuildEntryFilter(idArray, userId);
            return await DeleteListAsync(filter, entryType);

You should refer to the new pull request about how this works, basically it creates a single predicate and does a single search, so not only does this work now ( tested it) also it should work faster, because of a query for every id in the list, now it just uses one query for all of them

src/DataManagement/Dm.cs
```diff
@@ -0,0 +1,43 @@
...
+using ePiggyWeb.DataManagement.Goals;
+using ePiggyWeb.Utilities;
+
+namespace ePiggyWeb.DataManagement
+{
+    public class Dm
+    {
+        public IEntryManager Income { get; }
+        public IEntryManager Expenses { get; }
+        public IGoalManager Goals { get; }
```
Stokalas: What is this class?

ciliakas: This somehow got here by accident, was supposed to be stashed away :D It's data manager rework I had started working on

Stokalas (approved):
Seems ok now

PR #26, opened 2020-11-14
Goal database update

ciliakas:
Refactored Goal Database to be instance based instead of static and changed all of front end to use Goal Database instead of Data Manager ( for now at least, needed to test how everything works, if I decide to refactor data manager as well, then I'll reintegrate Data Manager)
Changes are pretty similar to the previous User database update, the old Goal Database class was renamed as GoalDatabaseOld (it's still in use by Data Manager).


Stokalas (approved):
Questions I would have are the same as I asked in previous pull request, since big part of this one is the refactoring of code in the same direction, so no pointing asking them again here. Did not notice any mistakes, aprroving

PR #27, opened 2020-11-14
Snow

Stokalas:
Just a small update/easter egg to bring christmas spirit.


Stokalas:
- Added links to footer icons
- Displays totals above the charts
- Removed background for Logo image, which is used in Index.html

ciliakas (approved):
This is really neat, I actually really enjoy the snow :D I wish they would stay on if you turn them on in every page, but that's really not worth the trouble to implement. The working links are a nice touch as well. 
Didn't go into the code that deeply, because as I see it's mostly html and css

Stokalas:
I wanted to keep it, but it is more difficult thant I expected since somewhy I can't save cookies from `_Layout`. So I thought I won't waste time investigating :D. Yup this branch is mainly design improvements, no logic or c#. 

**EDIT:** though I would like to know your oppinion on overall sums on chart pages.

ciliakas:
Ah, sorry, forgot to take a look at that.
I like the idea, my suggestion would be to if possible to display those numbers as currency, and maybe to display balance in the comparison chart?

PR #28, opened 2020-11-15
Display suggestions

Stokalas:
Creating this pull request a bit early just to allow to test calculations.

### DIFF
--- a/src/DataManagement/Saving/AlternativeSavingCalculator.cs
+++ b/src/DataManagement/Saving/AlternativeSavingCalculator.cs
@@ -27,11 +27,11 @@ namespace ePiggyWeb.DataManagement.Saving
 
             var enumCount = Enum.GetValues(typeof(Importance)).Length;
 
-            decimal[] arrayTemplate = new decimal[enumCount];
-            decimal[] sumsOfAmountByImportanceAdjusted = arrayTemplate;
-            decimal[] sumsOfAmountByImportanceDefault = arrayTemplate;
-            decimal[] averagesOfAmountByImportanceAdjusted = arrayTemplate;
-            decimal[] averagesOfAmountByImportanceDefault = arrayTemplate;
+            //decimal[] arrayTemplate = new decimal[enumCount];
+            decimal[] sumsOfAmountByImportanceAdjusted = new decimal[enumCount];
+            decimal[] sumsOfAmountByImportanceDefault = new decimal[enumCount];
+            decimal[] averagesOfAmountByImportanceAdjusted = new decimal[enumCount];
+            decimal[] averagesOfAmountByImportanceDefault = new decimal[enumCount];
             int[] entryAmounts = new int[enumCount];
 
             for (var i = enumCount; i > (int)Importance.Necessary; i--)
@@ -61,6 +61,12 @@ namespace ePiggyWeb.DataManagement.Saving
                 Debug.WriteLine("While");
                 for (var i = enumCount; i > (int)Importance.Necessary; i--)
                 {
+                    //bad, default and adjusted averages the same
+                    Debug.WriteLine(averagesOfAmountByImportanceAdjusted[i-1]);
+                    Debug.WriteLine(averagesOfAmountByImportanceDefault[i-1]);
+                    //seems 
+                    Debug.WriteLine(entryAmounts[i - 1]);
+                    Debug.WriteLine(i);
                     if (entryAmounts[i - 1] != 0)
                     {
                         averagesOfAmountByImportanceAdjusted[i - 1] = sumsOfAmountByImportanceAdjusted[i - 1] / entryAmounts[i - 1];
@@ -68,11 +74,11 @@ namespace ePiggyWeb.DataManagement.Saving
                     }
                     else
                     {
-                        averagesOfAmountByImportanceAdjusted[i - 1] = sumsOfAmountByImportanceAdjusted[i - 1];
-                        averagesOfAmountByImportanceDefault[i - 1] = sumsOfAmountByImportanceDefault[i - 1];
+                        averagesOfAmountByImportanceAdjusted[i - 1] = 0;
+                        averagesOfAmountByImportanceDefault[i - 1] = 0;
                     }
-                    
 
+                    Debug.WriteLine(approximateSavedAmount);
                     approximateSavedAmount += averagesOfAmountByImportanceDefault[i - 1] - averagesOfAmountByImportanceAdjusted[i - 1];
 
                     monthlySuggestions.Add(new SavingSuggestionByMonth(averagesOfAmountByImportanceAdjusted[i - 1], averagesOfAmountByImportanceDefault[i - 1], (Importance)i));
@@ -81,3 +87,4 @@ namespace ePiggyWeb.DataManagement.Saving
         }
     }
 }
+//Todo: fix while loop | add startingbalance into calculation | find out why for cycle loops in place
--- a/src/DataManagement/Saving/AlternativeSavingCalculator.cs
+++ b/src/DataManagement/Saving/AlternativeSavingCalculator.cs
@@ -55,7 +55,7 @@ namespace ePiggyWeb.DataManagement.Saving
 
                 }
             }
-            var approximateSavedAmount = 0M;
+            var approximateSavedAmount = startingBalance;
             while (goal.Amount > approximateSavedAmount)
             {
                 for (var i = enumCount; i > (int)Importance.Necessary; i--)
@@ -73,7 +73,8 @@ namespace ePiggyWeb.DataManagement.Saving
 
                     approximateSavedAmount += averagesOfAmountByImportanceDefault[i - 1] - averagesOfAmountByImportanceAdjusted[i - 1];
                 }
-
+                Debug.WriteLine(approximateSavedAmount);
+                Debug.WriteLine(timesToRepeatSaving);
                 timesToRepeatSaving++;              
             }
             for (var i = enumCount; i > (int)Importance.Necessary; i--)
--- a/src/DataManagement/Saving/AlternativeSavingCalculator.cs
+++ b/src/DataManagement/Saving/AlternativeSavingCalculator.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Diagnostics;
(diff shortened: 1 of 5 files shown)


src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
```
Stokalas: After deleting Debug code this should have been deleted too.

src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -8,42 +9,39 @@ namespace ePiggyWeb.DataManagement.Saving
...
-            decimal[] arrayTemplate = new decimal[enumCount];
-            decimal[] sumsOfAmountByImportanceAdjusted = arrayTemplate;
-            decimal[] sumsOfAmountByImportanceDefault = arrayTemplate;
-            decimal[] averagesOfAmountByImportanceAdjusted = arrayTemplate;
-            decimal[] averagesOfAmountByImportanceDefault = arrayTemplate;
+ 
+            decimal[] sumsOfAmountByImportanceAdjusted = new decimal[enumCount];
+            decimal[] sumsOfAmountByImportanceDefault = new decimal[enumCount];
+            decimal[] averagesOfAmountByImportanceAdjusted = new decimal[enumCount];
+            decimal[] averagesOfAmountByImportanceDefault = new decimal[enumCount];
```
Stokalas: all `decimal[]` should be changed to `var[]`

src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -54,19 +52,32 @@ public static class AlternativeSavingCalculator
...
             while (goal.Amount > approximateSavedAmount)
             {
                 for (var i = enumCount; i > (int)Importance.Necessary; i--)
                 {
-                    averagesOfAmountByImportanceAdjusted[i - 1] = sumsOfAmountByImportanceAdjusted[i - 1] / entryAmounts[i - 1];
-                    averagesOfAmountByImportanceDefault[i - 1] = sumsOfAmountByImportanceDefault[i - 1] / entryAmounts[i - 1];
+                    if (entryAmounts[i - 1] != 0)
+                    {
+                        averagesOfAmountByImportanceAdjusted[i - 1] = sumsOfAmountByImportanceAdjusted[i - 1] / entryAmounts[i - 1];
+                        averagesOfAmountByImportanceDefault[i - 1] = sumsOfAmountByImportanceDefault[i - 1] / entryAmounts[i - 1];
```
Stokalas, line 63: Wait does these arrays just hold the same value or something like that? (the what's the point of having two) And if same operations in two lines, maybe lambda expression could be used here, but not sure, since it is pretty simple equation.

src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -8,42 +9,39 @@ namespace ePiggyWeb.DataManagement.Saving
 {
     public static class AlternativeSavingCalculator
```
Stokalas, line 9: We should really think about leaving this class static of not.

Stokalas (commented):
Since I was responsible for UI I still reviewed back-end code, and left few notices. @DeividasLT5

src/Pages/SavingSuggestions.cshtml.cs
```diff
@@ -0,0 +1,82 @@
...
+            SetData();
+            return Page();
+        }
+
+        public void SetData()
+        {
+            UserId = int.Parse(User.FindFirst(ClaimTypes.Name).Value);
+            var dataManager = new DataManager(UserId);
+            Expenses = dataManager.Expenses.EntryList.GetFrom(StartDate).GetTo(EndDate);
+            Goal = dataManager.Goals.GoalList.FirstOrDefault(x => x.Id == Id);
```
ciliakas, line 67: We can add DI to this class when we pull our other pull requests

src/DataManagement/Saving/AlternativeSavingCalculator.cs
ciliakas: We should ask the lecturer what he thinks, and where should we use static classes in general.

src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -8,42 +9,39 @@ namespace ePiggyWeb.DataManagement.Saving
...
     {
-        //This might not be 1M in the future, that why we are keeping this
-        private static decimal SavingRatio { get; } = 1M;
         private static decimal RegularSavingValue { get; } = 0.25M;
         private static decimal MaximalSavingValue { get; } = 0.5M;
         private static decimal MinimalSavingValue { get; } = 0.1M;
 
-        public static void GetSuggestedExpensesOffers(IEntryList entryList, IGoal goal, IList<ISavingSuggestion> entrySuggestions,
+        public static int GetSuggestedExpensesOffers(IEntryList entryList, IGoal goal, IList<ISavingSuggestion> entrySuggestions,
             List<SavingSuggestionByMonth> monthlySuggestions, decimal startingBalance, SavingType savingType = SavingType.Regular)
```
ciliakas, line 16: To keep to our standard, maybe we should create an interface of SavingSuggestionByMonth, would also make this class less tightly coupled

src/DataManagement/Saving/AlternativeSavingCalculator.cs
ciliakas: actually just var, var[] would not compile :D

src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -8,42 +9,39 @@ namespace ePiggyWeb.DataManagement.Saving
...
                 {
                     var amountAfterSaving = savingType switch
                     {
-                        SavingType.Minimal => entry.Amount * SavingRatio * ratio * MinimalSavingValue,
-                        SavingType.Regular => entry.Amount * SavingRatio * ratio * RegularSavingValue,
-                        SavingType.Maximal => entry.Amount * SavingRatio * ratio * MaximalSavingValue,
+                        SavingType.Minimal => entry.Amount * ratio * MinimalSavingValue,
+                        SavingType.Regular => entry.Amount * ratio * RegularSavingValue,
+                        SavingType.Maximal => entry.Amount * ratio * MaximalSavingValue,
                         _ => throw new ArgumentOutOfRangeException()
```
ciliakas, line 44: Even though this exception can't really occur ( because SavingType will always be one of the three values we have), we probably should still catch this exception

src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -8,42 +9,39 @@ namespace ePiggyWeb.DataManagement.Saving
...
+                var expenses = entryList.GetBy((Importance)i);
                 var ratio = enumCount - i;
                 foreach (var entry in expenses)
                 {
                     var amountAfterSaving = savingType switch
                     {
-                        SavingType.Minimal => entry.Amount * SavingRatio * ratio * MinimalSavingValue,
-                        SavingType.Regular => entry.Amount * SavingRatio * ratio * RegularSavingValue,
-                        SavingType.Maximal => entry.Amount * SavingRatio * ratio * MaximalSavingValue,
+                        SavingType.Minimal => entry.Amount * ratio * MinimalSavingValue,
```
ciliakas: This whole methods is pretty large and quite hard to follow as it contains complex logic, it should probably by split into a few smaller methods if possible

src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -54,19 +52,32 @@ public static class AlternativeSavingCalculator
...
+                        averagesOfAmountByImportanceDefault[i - 1] = 0;
+                    }
 
                     approximateSavedAmount += averagesOfAmountByImportanceDefault[i - 1] - averagesOfAmountByImportanceAdjusted[i - 1];
-
-                    monthlySuggestions.Add(new SavingSuggestionByMonth(averagesOfAmountByImportanceAdjusted[i - 1], averagesOfAmountByImportanceDefault[i - 1], (Importance)i));
-                }                              
+                }
+                timesToRepeatSaving++;              
+            }
```
ciliakas, line 74: For example this looks like it could be it's own separate method

ciliakas (changes requested):
Seems to work after testing, but requires style adjustment and splitting into methods to make this quite complex logic easier to follow. Would be nice to maybe implement a requirement or two into this, I think it should be possible to add delegates or lambda expressions

src/DataManagement/Saving/AlternativeSavingCalculator.cs
Stokalas: Whoops, my bad :D

Stokalas:
Since it is merged into calculations branch and this was mainly front-end branch merging pull-request and expecting mentioned fixes in calculations branch.

src/DataManagement/Saving/AlternativeSavingCalculator.cs
DeividasLT5: I'll look into lambda expressions, but the arrays just hold the untouched values, values after they're adjusted and how many values there are in an importance, there's probably a better way to do this, but I went for simplicity and readability. Answer your question?

src/DataManagement/Saving/AlternativeSavingCalculator.cs
DeividasLT5: Essentially the whole calculation works with two big cycles, I can't really imagine a way to split those up in a way that results in LESS confusing logic, because you'd have to trace what values and how are they worked with in different methods. I would consider it if you provided a better example of what could be done

src/DataManagement/Saving/AlternativeSavingCalculator.cs
DeividasLT5: the problem is that the front would then need to call two seperate methods, no? where one is just a part of the whole, so I'm not sure about it. Or just seperate the method out and still call it, which, i feel like only adds to unreadability in this case.

PR #29, opened 2020-11-16
Predicate Builder

ciliakas:
Purpose: For more efficient deleting of many elements from database, instead of having many queries by every id, now we can call a single query and to look for all elements we want to find
ExpressionExtension:
Extension methods for Expression, used two add two predicates together into a single one, for only for And/Or logical operators
PredicateBuilder: Used for specific predicate building, at the moment it's used to create predicates for deleting or searching elements in database
This should fulfill at least in part the Generics requirement

### DIFF
--- a/src/Utilities/ExpressionExtension.cs
+++ b/src/Utilities/ExpressionExtension.cs
@@ -1,0 +1,25 @@
+﻿using System;
+using System.Linq.Expressions;
+
+namespace ePiggyWeb.Utilities
+{
+    public static class ExpressionExtension
+    {
+        public static Expression<Func<T, bool>> Or<T>(this Expression<Func<T, bool>> expression1, Expression<Func<T, bool>> expression2)
+        {
+            return expression1.Compose(Expression.Or, expression2);
+        }
+
+        public static Expression<Func<T, bool>> And<T>(this Expression<Func<T, bool>> expression1, Expression<Func<T, bool>> expression2)
+        {
+            return expression1.Compose(Expression.And, expression2);
+        }
+
+        private static Expression<Func<T, bool>> Compose<T>(this Expression<Func<T, bool>> expression1,
+            Func<Expression, Expression, BinaryExpression> logicalOperation, Expression<Func<T, bool>> expression2)
+        {
+            var invokedExpression = Expression.Invoke(expression2, expression1.Parameters);
+            return Expression.Lambda<Func<T, bool>>(logicalOperation.Invoke(expression1.Body, invokedExpression), expression1.Parameters);
+        }
+    }
+}
--- a/src/Utilities/PredicateBuilder.cs
+++ b/src/Utilities/PredicateBuilder.cs
@@ -1,0 +1,62 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Threading.Tasks;
+using ePiggyWeb.DataBase.Models;
+
+namespace ePiggyWeb.Utilities
+{
+    public static class PredicateBuilder
+    {
+        public static Expression<Func<IEntryModel, bool>> BuildEntryFilter(IEnumerable<int> idEnumerable, int userId)
+        {
+            if (idEnumerable is null)
+            {
+                return x => x.UserId == userId;
+            }
+
+            var idArray = idEnumerable as int[] ?? idEnumerable.ToArray();
+
+            if (!idArray.Any())
+            {
+                return x => x.UserId == userId;
+            }
+
+            Expression<Func<IEntryModel, bool>> filter = x => x.Id == idArray[0] && x.UserId == userId;
+
+            for (var i = 1; i <= idArray.Length; i++)
+            {
+                var i1 = i;
+                filter = filter.Or(x => x.Id == idArray[i1] && x.UserId == userId);
+            }
+
+            return filter;
+        }
+
+        public static Expression<Func<IGoalModel, bool>> BuildGoalFilter(IEnumerable<int> idEnumerable, int userId)
+        {
+            if (idEnumerable is null)
+            {
+                return x => x.UserId == userId;
+            }
+
+            var idArray = idEnumerable as int[] ?? idEnumerable.ToArray();
+
+            if (!idArray.Any())
+            {
+                return x => x.UserId == userId;
+            }
+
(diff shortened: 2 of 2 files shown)


src/Utilities/ExpressionExtension.cs
```diff
@@ -0,0 +1,25 @@
...
+        public static Expression<Func<T, bool>> And<T>(this Expression<Func<T, bool>> expression1, Expression<Func<T, bool>> expression2)
+        {
+            return expression1.Compose(Expression.And, expression2);
+        }
+
+        private static Expression<Func<T, bool>> Compose<T>(this Expression<Func<T, bool>> expression1,
+            Func<Expression, Expression, BinaryExpression> logicalOperation, Expression<Func<T, bool>> expression2)
+        {
+            var invokedExpression = Expression.Invoke(expression2, expression1.Parameters);
+            return Expression.Lambda<Func<T, bool>>(logicalOperation.Invoke(expression1.Body, invokedExpression), expression1.Parameters);
```
Stokalas, line 22: Source to read or just the difference between these? well what is different between first and second I understand, but what about third?

src/Utilities/PredicateBuilder.cs
```diff
@@ -0,0 +1,62 @@
...
+            }
+
+            var idArray = idEnumerable as int[] ?? idEnumerable.ToArray();
+
+            if (!idArray.Any())
+            {
+                return x => x.UserId == userId;
+            }
+
+            Expression<Func<IEntryModel, bool>> filter = x => x.Id == idArray[0] && x.UserId == userId;
```
Stokalas, line 26: isn't that a lambda expression ?

src/Utilities/PredicateBuilder.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            for (var i = 1; i <= idArray.Length; i++)
+            {
+                var i1 = i;
+                filter = filter.Or(x => x.Id == idArray[i1] && x.UserId == userId);
+            }
+
+            return filter;
+        }
```
Stokalas, line 35: For example, what exactly this method does?

Stokalas (commented):
Seems well written, just a bit confusing since I have no experience with such things. Left a few questions.

src/Utilities/PredicateBuilder.cs
ciliakas: It's exactly what it is, Expression stands for Lambda Expression, and you can store them as variable this way :)

src/Utilities/PredicateBuilder.cs
ciliakas: (Lines 14 to 24 just check if the id list is not null or not empty, if it is just returns a predicate/filter by user id)
So let's say we have an array of ids, like [1,2,3] and user id = 0
We go by each id in the array, and we add up predicate/filter/lambda expression.
The predicate we would have at the end would look like: 

      x => (x.Id == 1 && x.UserId == 0) || (x.Id == 2 && x.UserId == 0) || (x.Id == 3 && x.UserId == 0)

I hope this explains it :D

src/Utilities/PredicateBuilder.cs
Stokalas: But it does not fulfill the requirement?

src/Utilities/PredicateBuilder.cs
Stokalas: Oh, makes sense now. Well explained.

src/Utilities/PredicateBuilder.cs
ciliakas: Now that you say it, maybe it should actually. We need to ask the lecturer

Stokalas (approved):
When you explained these features make sense, approving.

src/Utilities/ExpressionExtension.cs
ciliakas: First thing to note, is what is the difference between Expression<Func<T1, T2>> and Func<T1,T2> is. The first is for simplicity sake, lets say a wrapped Func delegate, Expression allows building expression trees, as I understand, things like I'm making here. For example

     Expression<Func<int, bool>> expression = x => x < 2;
     Func<int, bool> function = expression.Compile();
(However, you can't create an Expression from a Func)

The compose method takes in the following parameters:

     this Expression<Func<T, bool>> expression1
     Expression<Func<T, bool>> expression2

are the two predicates we want to compose/fuse/merge into one.

    Func<Expression, Expression, BinaryExpression> logicalOperation

is a delegate, that takes two Expressions as parameters, and returns a Binary Expression (basically this is how I store the logical operation we want to merge by, like 'And' or 'Or').

            var invokedExpression = Expression.Invoke(expression2, expression1.Parameters);

this is where it gets a difficult to explain ( I don't understand this completely myself). Basically, invoking a Lambda Expression is applying an argument(s) ( in this case expression1 parameters) to a Lambda Expression ( in this case expression2). For example, invoking would look like

     Expression<Func<int, bool>> pred1 = x => x > 1000;
     Expression<Func<int, bool>> pred2 = y => y == 20;
     Expression.Invoke(pred1, pred1.Parameters)
     //Invoke(x => (x > 1000), y)      - this is how it looks printed out
Moving forward:

            return Expression.Lambda<Func<T, bool>>(logicalOperation.Invoke(expression1.Body, invokedExpression), expression1.Parameters);

So to start, Expression.Lambda creates a Lambda Expression, taking the first parameter as a body of the Lambda Expression, and later on parameters as the parameters of the Lambda Expression.
Then we invoke the logical Operation delegate (note that this is different invoke from the previous one), basically meaning:

     logicalOperation.Invoke(param1, param2) == Expression.And(param1, param2)

and then we pass expression1 parameters to this Lambda Expression.

(Why we have to use invoked expression in the logical operation, I'm not sure)

Keep in mind parts of the code I got from the internet, now spent an hour or so trying to understand fully how it works, so the explanation might not be the most accurate.
Sources:
https://docs.microsoft.com/en-us/dotnet/api/system.linq.expressions.expression.invoke?view=net-5.0
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/lambda-expressions
https://docs.microsoft.com/en-us/dotnet/api/system.linq.expressions.invocationexpression?view=net-5.0

ciliakas: I tried my best to explain this, but I still it's not very clear what this does :D

PR #31, opened 2020-11-16
Goal parsing

Stokalas:
Changed Internet parser a bit and added ability to parse goal by entered Title. Though parser is still static and I think it would be nice to use DI??
Also when we lost handler, HttpClient is created every time inside method which is bad practice.

Need suggestions, thoughts.

### DIFF
--- a/src/Pages/Goals.cshtml.cs
+++ b/src/Pages/Goals.cshtml.cs
@@ -1,10 +1,12 @@
 using System;
 using System.ComponentModel.DataAnnotations;
 using System.Security.Claims;
+using System.Threading.Tasks;
 using ePiggyWeb.DataBase;
 using ePiggyWeb.DataManagement;
 using ePiggyWeb.DataManagement.Entries;
 using ePiggyWeb.DataManagement.Goals;
+using ePiggyWeb.Utilities;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.RazorPages;
@@ -49,6 +51,19 @@ namespace ePiggyWeb.Pages
             }
             UserId = int.Parse(User.FindFirst(ClaimTypes.Name).Value);
             var temp = Goal.CreateLocalGoal(Title, Amount);
+            GoalDatabase.Create(temp, UserId);
+            return RedirectToPage("/goals");
+        }
+
+        public async Task<IActionResult> OnPostParseGoal()
+        {
+            if (string.IsNullOrEmpty(Title))
+            {
+                OnGet();
+                return Page();
+            }
+            UserId = int.Parse(User.FindFirst(ClaimTypes.Name).Value);
+            var temp = await InternetParser.ReadPriceFromCamel(Title);
             GoalDatabase.Create(temp, UserId);
             return RedirectToPage("/goals");
         }
--- a/src/Pages/Goals.cshtml.cs
+++ b/src/Pages/Goals.cshtml.cs
@@ -70,7 +70,7 @@ namespace ePiggyWeb.Pages
         {
             if (string.IsNullOrEmpty(Title))
             {
-                OnGet();
+                await OnGet();
                 return Page();
             }
             UserId = int.Parse(User.FindFirst(ClaimTypes.Name).Value);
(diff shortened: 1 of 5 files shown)


ciliakas:
Well I do know that we can use DI with Singleton scope ( basically one object for the whole server) which would kind of mimic static, but I don't know if that's the correct practice here. Probably would need to read up on how http clients are used in asp.net applications

src/Utilities/InternetParser.cs
```diff
@@ -1,32 +1,29 @@
...
     {
-        /*Example of calling method:
-         * Task.Run(() => InternetParser.GetHTMLAsync()).Wait();
-        */
-        private static readonly string ResourceDirectoryParsedGoal = Directory.GetParent(Environment.CurrentDirectory)
-                                                                         .Parent?.Parent?.FullName +
-                                                                     @"\resources\textData\parsedGoal.txt";
-        public static async Task ReadPriceFromCamel(string itemName)
+        private static readonly Func<int, string, bool> IsTooLong = (int x, string s) => s.Length > x;
+        private static readonly Func<decimal, decimal> ConvertToEur = (decimal price) => price * (decimal) 1.11;
```
ciliakas: Explicit declaration of types is unnecessary here, could just be (x, s) instead of (int x, string s)

ciliakas (approved):
Seems to be working fine, code looks fine as well, didn't go into the details of the functionality too much. Left a minor comment about unnecessary code

src/Pages/Goals.cshtml.cs
```diff
@@ -59,6 +66,19 @@ public async Task<IActionResult> OnPostNewGoal()
             return RedirectToPage("/goals");
         }
 
+        public async Task<IActionResult> OnPostParseGoal()
+        {
+            if (string.IsNullOrEmpty(Title))
+            {
+                OnGet();
```
ciliakas: I think this is missing an await

src/Utilities/InternetParser.cs
```diff
@@ -44,18 +43,31 @@ public static async Task ReadPriceFromCamel(string itemName)
...
+
+                var temp = Goal.CreateLocalGoal(name, decimalPrice);
+                return temp;
             }
-            catch (Exception ex)
+            catch(Exception ex)
             {
                 ExceptionHandler.Log(ex.ToString());
+                itemName = WebUtility.UrlDecode(itemName);
+                return Goal.CreateLocalGoal(itemName, 0);
```
ciliakas, line 71: Later on maybe alerting the user that the parser has failed with a simple dialog box would be a nice feature

ciliakas: So for example we could throw this further, and surround this parse statement with a try catch block in the front end and if we catch and exception we display the error to the user

src/Utilities/InternetParser.cs
```diff
@@ -1,32 +1,31 @@
...
-        {
-            var HttpClient = new HttpClient();
+        private readonly Func<int, string, bool> _isTooLong = (x, s) => s.Length > x;
+        private readonly Func<decimal, decimal> _convertToEur = (price) => price * (decimal) 1.11;
+        private readonly HttpClient _httpClient;
 
+        public InternetParser(HttpClient httpClient)
+        {
+            _httpClient = httpClient;
+        }
```
ciliakas: Nice idea to implement Func, but with the following code we could implement delegates here as well ( not that it makes any sense, but it would fulfill the requirement somewhat :D)

        private delegate decimal Cnv(decimal num);
        private readonly Cnv _cnv;
        public InternetParser(HttpClient httpClient)
        {
            _httpClient = httpClient;
            _cnv = _convertToEur.Invoke;
        }

ciliakas (changes requested):
Everything seems to work fine after testing, code looks good as well, the only thing that I think requires fixing is a missing await statement, and I left a few other suggestions as well

src/Utilities/InternetParser.cs
Stokalas: It's not that simple when post is called, because returning to the same page will move to the error page instead of alert, and the only simple solution I use to escape `POST` being called again on refresh is redirecting, so all exceptions will be lost :/

Stokalas: Same problem will not allow to bubble exeptions to income,expenses,goals. And the alert will only be shown when data is not loaded from db, but not when adding/deleting etc. since it always redirects to this page and it "forgets" everything :/

ciliakas: I see, well we will work with what we can, I don't think this will be too big of a problem for us, as long as we satisfy the requirement it's good

ciliakas (approved):
Issues fixed, approving

PR #32, opened 2020-11-17
Calculations

Stokalas:
This is the main branch of calculations that consists of back-end and already reviewed front-end.

Not solved problems in the back:
 - One big method should be divided into few smaller ones.
 - ? if this class should be static or not.

Backend code is done by @DeividasLT5 .
I've created this since we have less and less time to waste 🙂


src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -8,47 +8,75 @@ namespace ePiggyWeb.DataManagement.Saving
...
+                    }
+
+                    approximateSavedAmount += averagesOfAmountByImportanceDefault[i - 1] - averagesOfAmountByImportanceAdjusted[i - 1];
                 }
+                timesToRepeatSaving++;              
+            }
+            for (var i = enumCount; i > (int)Importance.Necessary; i--)
+            {
+                monthlySuggestions.Add(new SavingSuggestionByMonth(averagesOfAmountByImportanceAdjusted[i - 1], averagesOfAmountByImportanceDefault[i - 1], (Importance)i));
             }
```
Stokalas: I remember this produce list that repeatet itself quite a few times. How was this solved?

DeividasLT5: You're looking at the solution here, I simply ran another for loop, which has the sole purpose of generating the list, before this adding to list was above in the loop that repeated as many times, as months were needed to save

src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -6,20 +6,20 @@
 
 namespace ePiggyWeb.DataManagement.Saving
 {
-    public static class AlternativeSavingCalculator
+    public class AlternativeSavingCalculator
```
Stokalas, line 9: After answer from the lecturer I think we actually leave this class static, but the again what about if it uses threads? :D Or as I understand this one won't use it?

src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -6,20 +6,20 @@
...
         private static decimal MinimalSavingValue { get; } = 0.1M;
 
-        public static int GetSuggestedExpensesOffers(IEntryList entryList, IGoal goal, IList<ISavingSuggestion> entrySuggestions,
-            List<SavingSuggestionByMonth> monthlySuggestions, decimal startingBalance, SavingType savingType = SavingType.Regular)
+        public CalculationResults GetSuggestedExpensesOffers(IEntryList entryList, IGoal goal, decimal startingBalance, SavingType savingType = SavingType.Regular)
         {
-            entrySuggestions ??= new List<ISavingSuggestion>();
-
+            var entrySuggestions = new List<ISavingSuggestion>();
+            var monthlySuggestions = new List<SavingSuggestionByMonth>();
```
Stokalas: Nice one, since no point passing list as parameter since they are empty, considering that result can be returned in class or struct which holds all the required values.

src/DataManagement/Saving/CalculationResult.cs
```diff
@@ -0,0 +1,26 @@
...
+            MonthlySuggestions = monthlySuggestions;
+            TimesToRepeatSaving = timesToRepeatSaving;
+        }
+
+        public CalculationResults()
+        {
+            EntrySuggestions = new List<ISavingSuggestion>();
+            MonthlySuggestions = new List<SavingSuggestionByMonth>();
+            TimesToRepeatSaving = 0;
+        }
```
Stokalas: Where do you see empty constructor used? Since if it is empty there's really no point in displaying it etc.

src/DataManagement/Saving/TheadingCalculator.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Collections.Generic;
+using ePiggyWeb.DataManagement.Entries;
+using ePiggyWeb.DataManagement.Goals;
+using ePiggyWeb.Utilities;
+
+namespace ePiggyWeb.DataManagement.Saving
+{
+    public class ThreadingCalculator
```
Stokalas, line 11: As I understand this is the class which is supposed to work with the threads? Calling the alternateSavingsCalculator? Because then  alternateSavingsCalculator could be static since even if there's 3 threads they will be working with different lists.

Stokalas (commented):
I've reviewed back-end code, and left some questions, did not pay attention to front code since I'll review/fix when I'll be working on it myself :)

Stokalas:
I want to report a bug:
When trying to use ThreadingCalculator following exception happens:
image
In this line: 
image
And this method is called by: 
image
Probably something wrong with iteration of saving types.

src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -6,49 +6,90 @@
...
+                            {
+                                amountAfterSaving = entry.Amount;
+                            }
+                            break;
+                        case SavingType.Regular:
+                            amountAfterSaving = entry.Amount * ratio * RegularSavingValue;
+                            break;
+                        case SavingType.Maximal:
+                            amountAfterSaving = entry.Amount * ratio * MaximalSavingValue;                            
+                            break;
```
Stokalas: I think there should still be default case which throws the exceptions, just in case :D

Stokalas (commented):
Seems to be working, just the questions from earlier and this one

src/DataManagement/Saving/AlternativeSavingCalculator.cs
DeividasLT5: Well the asynchronous part uses this call, so i made it non-static just in case, though I'm definitely willing to change it back, if that's the better option.

src/DataManagement/Saving/CalculationResult.cs
DeividasLT5: I think I used it in the SendData method in the front, as a quick work-around to test with compiling, or something, though it doesn't seem to have references anymore, so in such a case should i just delete it?

src/DataManagement/Saving/TheadingCalculator.cs
DeividasLT5: What's the advantage to using static, anyway, could you remind/tell me? At least at the moment it doesn't seem like an issue to me

src/DataManagement/Saving/AlternativeSavingCalculator.cs
DeividasLT5: Hmm, well, can't hurt I guess :D

src/DataManagement/Saving/TheadingCalculator.cs
DeividasLT5: Also, will they be working with different lists? :D i'm not sure about that. probably safer to just leave it like that, no?

src/DataManagement/Saving/CalculationResult.cs
Stokalas: I think best practice would be to delete it

src/DataManagement/Saving/AlternativeSavingCalculator.cs
DeividasLT5: I was slightly annoyed at this for quite some time, finally made a result class :D

src/DataManagement/Saving/AlternativeSavingCalculator.cs
Stokalas: Nice that you've implemented that

src/DataManagement/Saving/TheadingCalculator.cs
Stokalas: I think we can leave it as it is

PR #33, opened 2020-11-17
Config requirement

ciliakas:
Implemented options pattern for email sender and removed redundant code in startup.

### DIFF
--- a/src/Startup.cs
+++ b/src/Startup.cs
@@ -37,22 +37,10 @@ namespace ePiggyWeb
 
             services.AddDbContext<PiggyDbContext>(options =>
                 options.UseSqlServer(Configuration.GetConnectionString("DefaultConnection")));
-            services.AddScoped(provider =>
-            {
-                var db = provider.GetService<PiggyDbContext>();
-                return new UserDatabase(db);
-            });
-            services.AddScoped(provider =>
-            {
-                var db = provider.GetService<PiggyDbContext>();
-                return new EntryDatabase(db);
-            });
-            services.AddScoped(provider =>
-            {
-                var db = provider.GetService<PiggyDbContext>();
-                return new GoalDatabase(db);
-            });
-            services.AddScoped<EmailSender>();
+            services.AddScoped<UserDatabase>();
+            services.AddScoped<EntryDatabase>();
+            services.AddScoped<GoalDatabase>();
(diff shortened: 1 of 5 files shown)


src/Startup.cs
```diff
@@ -37,22 +37,10 @@ public void ConfigureServices(IServiceCollection services)
...
-            services.AddScoped(provider =>
-            {
-                var db = provider.GetService<PiggyDbContext>();
-                return new GoalDatabase(db);
-            });
-            services.AddScoped<EmailSender>();
+            services.AddScoped<UserDatabase>();
+            services.AddScoped<EntryDatabase>();
+            services.AddScoped<GoalDatabase>();
+            services.Configure<EmailSender>(options => Configuration.GetSection("Email").Bind(options));
```
Stokalas, line 44: So he know which data to read as options by section? And then assigns it to same name properties?

src/appsettings.json
```diff
@@ -6,5 +6,18 @@
...
+    "RecoveryMessageSubject": "Password Recovery",
+    "RecoveryMessageBody": "Your password recovery code is: ",
+    "GreetingMessageSubject": "Welcome to ePiggy!",
+    "GreetingMessageBody": "Thank you for joining ePiggy! Your wallet thanks you as well!",
+    "FarewellMessageSubject": "Farewell from the ePiggy team",
+    "FarewellMessageBody": "We're sorry to see you go, please leave your feedback of our app.",
+    "MaxRandValue": 999999,
+    "SmtpEmail": "smtp.gmail.com",
+    "Port": 587
+  },
```
Stokalas, line 21: And as I understand when we read these as options we can change them during runtime and changes will be applied without restarting?

Stokalas (approved):
Seems well done. Nice to see that requirements are coming to an end bit by bit.
Left a few questions about how it works, as always :)

src/Startup.cs
ciliakas: Basically that's how it works, you can read more here: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-5.0

src/appsettings.json
ciliakas: I think that's how it works, but I tried for a bit and couldn't get it to work, so maybe I'm misunderstanding it

PR #34, opened 2020-11-18
Events

ciliakas:
Added very basic events, both custom and standard to User Database, they can be subscribed to by any class that uses it.
Also removed Data Manager from Goal.cshtml.cs


src/Pages/Goals.cshtml.cs
```diff
@@ -29,17 +30,20 @@ public class GoalsModel : PageModel
...
 
         public async Task OnGet()
         {
             UserId = int.Parse(User.FindFirst(ClaimTypes.Name).Value);
-            var dataManager = new DataManager(UserId);
             Goals = await GoalDatabase.ReadListAsync(UserId);
-            Savings = dataManager.Income.EntryList.GetSum() - dataManager.Expenses.EntryList.GetSum();
+            var incomes = await EntryDatabase.ReadListAsync(UserId, EntryType.Income);
+            var expenses = await EntryDatabase.ReadListAsync(UserId, EntryType.Expense);
+            Savings = incomes.GetSum() - expenses.GetSum();
```
Stokalas, line 51: This is where the dataManager was removed in goals :D

src/Pages/Entry.cshtml.cs
```diff
@@ -14,17 +15,28 @@ namespace ePiggyWeb.Pages
...
+            GoalDatabase = goalDatabase;
+            UserDatabase = userDatabase;
+            EmailSender = emailSender;
+            UserDatabase.Deleted += OnDeleteUser;
+        }
+
+        private async void OnDeleteUser(object sender, UserModel user)
+        {
+            await EmailSender.SendFarewellEmailAsync(user.Email);
         }
```
Stokalas, line 35: As I undertand we need delete user button somewhere?

src/Pages/Login.cshtml.cs
```diff
@@ -27,9 +29,17 @@ public class LoginModel : PageModel
...
+        public LoginModel(UserDatabase userDatabase, EmailSender emailSender)
         {
             UserDatabase = userDatabase;
+            EmailSender = emailSender;
+            UserDatabase.Registered += OnRegister;
+        }
+
+        private async void OnRegister(object sender, UserModel user)
+        {
+            await EmailSender.SendGreetingEmailAsync(user.Email);
```
Stokalas: Why is greetings email in login? :D

src/Pages/Register.cshtml.cs
```diff
@@ -36,6 +37,12 @@ public RegisterModel(UserDatabase userDatabase, EmailSender emailSender)
         {
             UserDatabase = userDatabase;
             EmailSender = emailSender;
+            UserDatabase.Registered += OnRegister;
+        }
+
+        private async void OnRegister(object sender, UserModel user)
+        {
+            await EmailSender.SendGreetingEmailAsync(user.Email);
         }
```
Stokalas, line 47: Seems neat to send the greetings email in the first place, second nice that events were used :D

Stokalas (commented):
I did not test whether or not this works, but I'm happy that events were used, left few questions, will approve when answered. 🙂

src/Pages/Login.cshtml.cs
ciliakas: I have no idea :D Changed it to be On Login, even though it doesn't do anything, just to show that we could do something :D

src/Pages/Entry.cshtml.cs
ciliakas: That would be nice to have, maybe on the header or footer for now, just to show that we have this sort of functionality

src/Pages/Goals.cshtml.cs
ciliakas: Now it makes sense :D

Stokalas (approved):

PR #35, opened 2020-11-18
Additional lazy init

ciliakas:
Added a few lazy initiliazations to hopefully better fulfill the requirement

### DIFF
--- a/src/Pages/ForgotPassword.cshtml.cs
+++ b/src/Pages/ForgotPassword.cshtml.cs
@@ -36,7 +36,7 @@ namespace ePiggyWeb.Pages
         private Lazy<EmailSender> EmailSender { get; }
         private UserDatabase UserDatabase { get; }
         private IConfiguration Configuration { get; }
-        public ForgotPasswordModel(UserDatabase userDatabase, IOptions<EmailSender> emailSenderSettings, IConfiguration configuration)
+        public ForgotPasswordModel(UserDatabase userDatabase, IConfiguration configuration)
         {
             UserDatabase = userDatabase;
             Configuration = configuration;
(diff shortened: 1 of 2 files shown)


src/Utilities/EmailSender.cs
```diff
@@ -8,6 +8,7 @@ namespace ePiggyWeb.Utilities
 {
     public class EmailSender
     {
+        public const string Email = "Email";
```
Stokalas, line 11: Where is this used?

src/Pages/ForgotPassword.cshtml.cs
```diff
@@ -52,7 +55,9 @@ public async Task<IActionResult> OnGet()
                 return Page();
             }
             Email = Request.Cookies["Email"];
-            var recoveryCode = (await EmailSender.SendRecoveryCodeAsync(Email)).ToString();
+
+            Configuration.GetSection(Utilities.EmailSender.Email).Bind(EmailSender.Value);
```
Stokalas, line 59: Can you explain what this line does?

Stokalas (approved):
Nothing much to say, since as I understand you simply added lazy init to already existing code. Just twu questions

src/Utilities/EmailSender.cs
ciliakas: This is used in the place where you asked your other question, this string is used to access the data from app settings through configuration

src/Pages/ForgotPassword.cshtml.cs
ciliakas: So I'm trying out options pattern a bit more, and what this does is takes a section from app settings (GetSection(Utilities.EmailSender.Email)) where the section name is "Email" (from the const string in above question) and then binds it to the properties of the EmailSender object.
Basically this gets data from app settings and injects it into a object

PR #36, opened 2020-11-20
Semi Random List Generation

ciliakas:
Added a static method to generate semi random lists, using the options pattern to get hardcoded titles from appsettings and creating some random numbers for the prices. Use example is shown in Entry.cshmtl.cs

### DIFF
--- a/src/DataManagement/Entries/EntryList.cs
+++ b/src/DataManagement/Entries/EntryList.cs
@@ -3,6 +3,9 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using ePiggyWeb.Utilities;
+using System.Configuration;
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.Extensions.Configuration;
 
 namespace ePiggyWeb.DataManagement.Entries
 {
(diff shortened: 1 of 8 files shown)


src/DataManagement/Entries/EntryList.cs
```diff
@@ -79,5 +82,28 @@ public override string ToString()
 
             return sb.ToString();
         }
+
+        [SuppressMessage("ReSharper", "PossibleNullReferenceException")]
+        public static IEntryList RandomList(IConfiguration configuration, EntryType entryType)
+        {
+            IEntryList list = new EntryList(entryType);
+            var section = configuration.GetSection(entryType.ToString());
```
Stokalas, line 90: what is this line ? :D

src/Pages/Login.cshtml.cs
```diff
@@ -30,10 +31,10 @@ public class LoginModel : PageModel
 
         private UserDatabase UserDatabase { get; }
         private EmailSender EmailSender { get; }
-        public LoginModel(UserDatabase userDatabase, EmailSender emailSender)
+        public LoginModel(UserDatabase userDatabase, IOptions<EmailSender> emailSenderSettings)
         {
             UserDatabase = userDatabase;
-            EmailSender = emailSender;
+            EmailSender = emailSenderSettings.Value;
```
Stokalas: Is this some kind of fix related that it don't works for me in dev? :D

Stokalas (approved):
Will test it when implementing, seems nice

src/Pages/Login.cshtml.cs
ciliakas: Yeah, I've fixed in dev now as well :D

src/DataManagement/Entries/EntryList.cs
ciliakas: The configuration object is used to access the appsettings file, from which we take the section of either 'Income' or 'Expense' and use the values in it to set the names of the entries in the list

Stokalas:
I've checked in dev and now Email thing works, nice ad quick fix! :D

Stokalas:
Also I've tested it inside my exceptions branch and it seems to be working well.

PR #37, opened 2020-11-21
Exceptions

Stokalas:
Files that are related to calculations should not be paid attention to, it was merge into here just for testing purposes :)
 - Added exception page, to which user is redirected when exception not caught
 - Exception bubbling and catching in page handlers, in some cases  filling list with placeholder data displaying it with alert;
 - A lot of files changed since old Exception handling was removed, and some files commented out :)
 - **_After reviewing this hughmongous pull request reward yourself with cookie or cake etc._** 🙂


src/DataBase/EntryDatabase.cs
```diff
@@ -92,19 +95,21 @@ public async Task<bool> UpdateSingleAsync(int id, int userId, IEntry updatedEntr
...
+                    x => x.Id == id && x.UserId == userId),
                 _ => await Database.Expenses.FirstOrDefaultAsync(x => x.Id == id && x.UserId == userId)
             };
 
+
             if (temp is null)
             {
-                ExceptionHandler.Log("Couldn't find entry id: " + id + " in database");
-                return false;
+                throw new Exception("Couldn't find entry id: " + id + " in database");
```
ciliakas, line 106: Good idea to move the error handling to the classes that actually use this class

src/Pages/ChangePassword.cshtml.cs
```diff
@@ -3,13 +3,18 @@
...
+using ePiggyWeb.Pages.ErrorPages;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.RazorPages;
+using Microsoft.Extensions.Logging;
 
 namespace ePiggyWeb.Pages
 {
     public class ChangePasswordModel : PageModel
     {
+        private readonly ILogger<ChangePasswordModel> _logger;
```
ciliakas, line 19: I've seen this ILogger a bunch of times in tutorials and whatnot, what is it for exactly? To log errors or general information?

src/Pages/ChangePassword.cshtml.cs
```diff
@@ -40,21 +47,30 @@ public IActionResult OnGet()
...
+                    ErrorMessage = "Passwords did not match!";
+                    return Page();
+                }
             }
-            else
+            catch (Exception ex)
             {
-                ErrorMessage = "Passwords did not match!";
+                _logger.LogInformation(ex.ToString());
+                WasException = true;
```
ciliakas: What is this property for?

src/Pages/ComparisonGraph.cshtml.cs
```diff
@@ -56,9 +59,22 @@ public async Task<IActionResult> OnGetFilter(DateTime startDate, DateTime endDat
...
+                    .GetTo(EndDate).GetSum();
+                Income = (await EntryDatabase.ReadListAsync(UserId, EntryType.Income)).GetFrom(StartDate).GetTo(EndDate)
+                    .GetSum();
+            }
+            catch (Exception ex)
+            {
+                _logger.LogInformation(ex.ToString());
+                WasException = true;
+                Expenses = 1000;
+                Income = 1000;
```
ciliakas: I think you could use the randomly generated lists here for the amount, even though it really doesn't make that much of a difference :D

src/Pages/ChangePassword.cshtml.cs
ciliakas: Because I see that in other pages you use this in the cshtml part, but I can't seem to find it in the html for this page

src/Pages/ErrorPages/Error.cshtml.cs
```diff
@@ -0,0 +1,31 @@
+using System.Diagnostics;
+using System.IO;
+using Microsoft.AspNetCore.Diagnostics;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.RazorPages;
+using Microsoft.Extensions.Logging;
+
+namespace ePiggyWeb.Pages.ErrorPages
+{
+    [ResponseCache(Duration = 0, Location = ResponseCacheLocation.None, NoStore = true)]
```
ciliakas, line 10: What does this do?

src/Pages/ErrorPages/Error.cshtml.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            _logger = logger;
+        }
+        public void OnGet()
+        {
+            RequestId = Activity.Current?.Id ?? HttpContext.TraceIdentifier;
+
+            var exceptionHandlerPathFeature =
+                HttpContext.Features.Get<IExceptionHandlerPathFeature>();
+            _logger.LogInformation(exceptionHandlerPathFeature.ToString());
```
ciliakas, line 28: Could you explain what does this do?

src/Startup.cs
```diff
@@ -46,16 +47,18 @@ public void ConfigureServices(IServiceCollection services)
...
-                app.UseDeveloperExceptionPage();
+                app.UseExceptionHandler("/ErrorPages/Error");
+                //app.UseDeveloperExceptionPage();
             }
             else
             {
-                app.UseExceptionHandler("/Error");
-                // The default HSTS value is 30 days. You may want to change this for production scenarios, see https://aka.ms/aspnetcore-hsts.
+                app.UseExceptionHandler("/ErrorPages/Error");
                 app.UseHsts();
```
ciliakas, line 62: Could you explain what does this do? I understand that it sets different settings if it's development environment or not, but the details I don't know about

src/ePiggyWeb.csproj
```diff
@@ -14,6 +14,11 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Serilog.Extensions.Logging.File" Version="2.0.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="Logs\" />
```
ciliakas, line 21: I assume this is used for logging?

ciliakas (commented):
Good job, I've tested this out a bit and it seems to work as intended, but I left a few questions about thing I don't understand about and I'll approve afterwards

src/Pages/ChangePassword.cshtml.cs
Stokalas: Yup it is to log exceptions, as I understand by default it logs to console or something, but I use package that automatically logs to file with today's date in its name.
[Tutorial I used](https://nblumhardt.com/2016/10/aspnet-core-file-logger/)

src/Pages/ChangePassword.cshtml.cs
Stokalas: If you mean `wasException` - it stores bool if exception happened and then in the actual csthml (not cshtml.cs) ir allows to load different kind of data using if's (for example not to show date filters etc.)

src/Pages/ComparisonGraph.cshtml.cs
Stokalas: I'm not sure about using lists since it would be more time costly and in this page separate entries does not matter that much. Though I would be happy to use random numbers in range 500-5000 here, if time allows me :D

src/ePiggyWeb.csproj
Stokalas: Yup it is package that allows to log exceptions into text file.

src/Startup.cs
Stokalas: Basically it sets that when uncatched exception happens it redirects to certain page. DeveloperExceptionPage is one I think you're familiar with (where all stack traces printed, one that was shown during all exceptions). And since we're testing inside Visual studio it uses developer enviroment, so I had to comment out. What is inside the if - what happens in dev enviroment, what is inside else - what would be displayed in production version (I use same thing for both).

_Hsts is a security feature to force SSL. It has nothing to do with exceptions._ - [Source](https://stackoverflow.com/questions/51093139/what-is-the-difference-between-app-usehsts-and-app-useexceptionhandler)

src/Pages/ErrorPages/Error.cshtml.cs
Stokalas: `var exceptionHandlerPathFeature =
                HttpContext.Features.Get<IExceptionHandlerPathFeature>();` - somehow gets the info about exception (not sure how), and
` _logger.LogInformation(exceptionHandlerPathFeature.ToString());` - logs it to file (I'm logging when we get into expection page as well as when handled exceptions)
This whole  `Error.cshtml.cs` was very much "_inspired_" by info on the internet, so I'm not sure about 100% how it works :D

src/Pages/ErrorPages/Error.cshtml.cs
Stokalas: **Response caching reduces the number of requests a client or proxy makes to a web server. Response caching also reduces the amount of work the web server performs to generate a response. Response caching is controlled by headers that specify how you want client, proxy, and middleware to cache responses.** 
- The ResponseCacheAttribute specifies the parameters necessary for setting appropriate headers in response caching.
 - To enable caching, Duration must be set to a positive value 
- NoStore overrides most of the other properties. When this property is set to true, the Cache-Control header is set to no-store. NoStore is typically set to true for error pages. 
 - Setting Location to None sets both Cache-Control and Pragma headers to no-cache.
I'm myself do not understand this completely but it just sets everything up so it would be handled like error page.
[Source](https://docs.microsoft.com/en-us/aspnet/core/performance/caching/response?view=aspnetcore-5.0)

src/Pages/ChangePassword.cshtml.cs
ciliakas: Sorry, I've mixed up

      Because I see that in other pages you use this in the cshtml part, but I can't seem to find it in the html for this page

this comment was meant to be here :D

src/Pages/ComparisonGraph.cshtml.cs
ciliakas: Makes sense :D

src/Pages/ErrorPages/Error.cshtml.cs
ciliakas: Aaaah I see, happens to me all too often to :D Could you explain a bit what does line 24 do as well?

ciliakas (approved):
Approving, just left a few more questions to your answers :)

src/Pages/ErrorPages/Error.cshtml.cs
Stokalas: I'll just add [stack overflow question](https://stackoverflow.com/questions/54906115/activity-id-in-asp-net-core-2-application)  regarding this

src/Pages/ChangePassword.cshtml.cs
Stokalas: Oh, I actually undestand what you've meant now, yup it got here by mistaking (by copying code). Since I only intended to log exceptions here regarding accessing database and not to redirect to error page, so bool is not needed here. WIll fix, thanks!

PR #38, opened 2020-11-21
Formatting in Income and Expenses Pages

Stokalas:
- Using NumberFormatter to Format Currency insteam of formating decimal in string Culture info.


ciliakas (changes requested):
Good idea, but I think we should implement this in more places, because for example the totals shown in Incomes/Expenses are still displayed as previously

ciliakas (approved):

PR #39, opened 2020-11-21
Delete user

Stokalas:
- Ability to delete account inside change password page; (with emails)
 - Different error messages when wrong login info entered

### DIFF
--- a/src/Pages/ChangePassword.cshtml.cs
+++ b/src/Pages/ChangePassword.cshtml.cs
@@ -3,8 +3,12 @@ using System.ComponentModel.DataAnnotations;
 using System.Security.Claims;
 using System.Threading.Tasks;
 using ePiggyWeb.DataBase;
+using ePiggyWeb.DataBase.Models;
+using ePiggyWeb.Utilities;
+using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.RazorPages;
+using Microsoft.Extensions.Options;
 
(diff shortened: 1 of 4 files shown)


src/Pages/ChangePassword.cshtml.cs
```diff
@@ -56,5 +64,17 @@ public async Task<IActionResult> OnPost()
                 return Page();
             }
         }
+
+        public async Task<IActionResult> OnPostDeleteAccount()
+        {
+            await HttpContext.SignOutAsync();
+            await UserDatabase.DeleteUserAsync(User.FindFirst(ClaimTypes.Email).Value);
+            return Redirect("/index");
+        }
```
ciliakas, line 73: Maybe we should delete all data associated with the user here as well so we don't clog up the database with data that does belong to inactive users

ciliakas (commented):
Good update, like the new error message, left a suggestion in the comments

src/Pages/ChangePassword.cshtml.cs
Stokalas: I think more logical approach would so the method `DeleteUserAsync` would call such method to delete data, since it would protect from putting more logic inside front.

ciliakas: That makes a lot more sense, I've implemented this into this branch, and I think this is ready to be approved

Stokalas: Glad to hear. I guess, yeah, since when I tested it worked I think I'll merge this one.

PR #40, opened 2020-11-21
Completed app?

Stokalas:
**Merge this after Exceptions and Calculations branches merged**
When those two will be merged I think the app will be +/- completed so we can merge everything into master and then do small fixes by "checkout'ing" from main instead of using dev.
Would be nice to skim through atleast some of the code again, and most importantly test in search of bugs/glitches.

PS it's kinda cool that this is 40th pull request :D


ciliakas:
I agree, after those two pull requests it we should pull everything into Main and leave it as is more or less and go on to the next one :D

PR #43, opened 2020-11-29
Mobileindex

Stokalas:
My first part of reworks which touches index page, layout (navbar and footer) and login/register pages. To work better on mobile devices.
(I'm testing everything by making browser's width as small as I can by dragging, also sometimes by pressing F12 and selecting mobile devices, though as I understand when "simulating touch" option selected it displays for wide screen so not sure how that thing works)

**Changes:**
 -  Removed unused css code
 -  Fixed logo in navbar going bigger when small screen
 -  Made login to fit whole screen width when small screen
 -  Added login button to index page
 -  Title image and testimonial text resizing on mobile
 -  Removed "Changed Password" button, not that page opens when clicked on email.
 -  Removed "I'm feeling christmas!" button (planning to move it to changePassword.cshtml as a account setting)


src/wwwroot/css/styles.css
```diff
@@ -39,11 +41,13 @@ body {
...
-    left: 5%;
-    bottom: 63%;
-}
 
+.title-image {  
+    max-width: 100%;
+    width: 30%;
+    height: auto;
+}
+}
```
ciliakas: One bracket too many here :)

src/Pages/Index.cshtml
```diff
@@ -16,7 +16,7 @@
 }
 
 <div class="container index-Top">
-    <img width="350px" src="images/mainLogo.PNG" alt="" class="title-image mt-2">
+    <img src="images/mainLogo.PNG" class="title-image mt-2">
```
ciliakas, line 19: Maybe in the future we could add alt text so that the website would become more accessible with people with visual impairments or for those who have bad internet connections

ciliakas (commented):
I wish I could comment more on this request, however it's quite hard for me to understand how html works just yet ( I'll sit down soon to get more acquainted with our front end code)
The responsive design looks great, the change in header seems like a great job. Other than that,I noticed a couple of interesting things, and I'm not sure if this is how the site is supposed to look:

bug1
bug2

Stokalas:
Login form should expand when smaller screen, I'm not sure why it is not doing that for you :D. Since this is updating CSS quite a lot each time before testing everyone should do CTRL + F5 to reload CSS file. And I did not touch change password page yet, only Index, Login and Register pages

src/Pages/Index.cshtml
Stokalas: Well it would be nice small touch, but since it is not as important for us now, maybe I'll do this later :)

ciliakas:
Well it does work when I make the page small enough, but there is some screen sizes ( like the pictures I sent) that don't seem to fit properly

ciliakas (approved):
After going with you trough the code, I understand what's happening here much better :D

PR #44, opened 2020-11-30
Mobilegoals

Stokalas:
- Changed "cards" in which goals are displayed
 - Desktop - 3; tablet - 2; mobile - 1 cards should be displayed.
 - Snow button in account settings
--------------------------------------------
 - All 3 chart pages mobile responsive
--------------------------------------------
_Merging into mobileindex since this branch was derived from it_


ciliakas (changes requested):
Good work going through the responsive part so quickly!  Looks like we are having some great progress. I just noticed a few ( what I think are) bugs while testing.

bug1

Something happened to sign in form, it was resizing fine previously. Also I couldn't find a way to log in or sign out when in mobile mode ( only register)

bug4

I know this is not important, but the picture are very tiny :D 

bug2

bug3

I don't know if this is difficult to fix, but the graphs don't scale after they are generated ( of course this is unlikely to be spotted by most users). The graphs itself look kind of small when in mobile mode, with a lot of space in between them, and the pie chart is a bit of center it seems.

All in all good work, I think this just needs a little bit of adjusment

Stokalas:
That's super weird since screenshots you've posted are from code before (one sign is that chart legend is now at the top of the graph.
image
 I'm not sure how and why but I haven't touched the login stuff, and it exists for me 
image

Also login form resizes as it should. One thing I can agree to is small image in testimonials but that will be fixed in mobileindex branch.

ciliakas:
I feel stupid now, I either messed up with git or didn't refresh the page properly, either way everything looks great now!

ciliakas (approved):
Looked at the code, as far as I understand it looks fine to me. Good job

PR #45, opened 2020-12-01
Recurring Fix

ciliakas:
Fixed the error when adding a new recurring entry to the current month, fixed editing entries into recurring entries, and added auto entry update on every read ( I know this is inefficient, but due to time constraints I think this will do for now)

### DIFF
--- a/src/DataBase/EntryDatabase.cs
+++ b/src/DataBase/EntryDatabase.cs
@@ -82,28 +82,21 @@ namespace ePiggyWeb.DataBase
                 return false;
             }
 
-            if (updatedEntry.Recurring)
-            {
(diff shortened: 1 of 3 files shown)


MatasLa (approved):
Tested it and everything seems to work fine now. However have no comments on code yet. (Still trying to catch up on it)

Stokalas (approved):
It does not add entry to the next month when entered monthly as it did previously, atleast as far as I tested. 
And as I understand code is changed so that it does not update entries when added/edited, but now it updates monthly stuff when reading data?

ciliakas:
> 
> 
> It does not add entry to the next month when entered monthly as it did previously, atleast as far as I tested.
> And as I understand code is changed so that it does not update entries when added/edited, but now it updates monthly stuff when reading data?

Now it automatically updates both when adding and editing, and when reading

PR #46, opened 2020-12-01
Filter cookies

ciliakas:
Added cookies to save filter dates. Cookies last 15 minutes, and are removed if logging out.

### DIFF
--- a/src/Utilities/TimeManager.cs
+++ b/src/Utilities/TimeManager.cs
@@ -54,7 +54,7 @@ namespace ePiggyWeb.Utilities
                 pageEndDate = endDate;
                 var cookieOptions = new CookieOptions
                 {
-                    Expires = DateTimeOffset.UtcNow.AddMinutes(15)
+                    Expires = DateTimeOffset.UtcNow.AddHours(4)
                 };
                 response.Cookies.Append("StartDate", startDate.ToString(CultureInfo.InvariantCulture), cookieOptions);
                 response.Cookies.Append("EndDate", endDate.ToString(CultureInfo.InvariantCulture), cookieOptions);
--- a/src/Utilities/TimeManager.cs
+++ b/src/Utilities/TimeManager.cs
@@ -1,10 +1,62 @@
 ﻿using System;
+using System.Globalization;
+using Microsoft.AspNetCore.Http;
 
 namespace ePiggyWeb.Utilities
 {
     public static class TimeManager
     {
         public static DateTime OneMonthAhead { get; }= GetEndOfTheMonth(DateTime.Today.AddMonths(1));
+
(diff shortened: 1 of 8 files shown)


src/Utilities/TimeManager.cs
```diff
@@ -1,11 +1,63 @@
...
+using System.Globalization;
+using Microsoft.AspNetCore.Http;
 
 namespace ePiggyWeb.Utilities
 {
     public static class TimeManager
     {
         public static DateTime OneMonthAhead { get; }= GetEndOfTheMonth(DateTime.Today.AddMonths(1));
 
+        public static void GetDate(HttpRequest request, out DateTime pageStartDate, out DateTime pageEndDate)
```
Stokalas, line 16: these methods seems nice, since they moved some logic away from front.

src/Utilities/TimeManager.cs
```diff
@@ -1,11 +1,63 @@
...
+            }
+            else
+            {
+                pageStartDate = startDate;
+                pageEndDate = endDate;
+                response.Cookies.Delete("StartDate");
+                response.Cookies.Delete("EndDate");
+                var cookieOptions = new CookieOptions
+                {
+                    Expires = DateTimeOffset.UtcNow.AddMinutes(15)
```
Stokalas: Not sure about the option, since this allows filters to be saved only for 15 mins. Was this intended?

Stokalas (approved):
Seems to be working as I've tested. I really like that it cleared front code quite a bit. Just not sure about cookie option to stay alive for 15 mins. But it's a minor detail.

src/Utilities/TimeManager.cs
ciliakas: I really don't know how long these filtering options should last, so now they just last enough for active testing. I'm open to suggestions on how long they should stay :)

Stokalas: I guess until logging off? Since they can be changed and would be nice if stayed consistent through whole "workflow of the user". And then reset until next time.

ciliakas: I'm not sure how to delete the cookies on the automatic sign out ( when the authentication cookies expire), but I can make the filter cookies last a long time and have them removed when manually signing out

ciliakas: I tried to make these cookies sliding just like the authorization cookies, however that proved quite difficult to implement and I don't think it's worth the struggle :D

Stokalas: I think then the most logical option is like you've said, to give them lifetime of around 3-5 hours and leave them as it is.

ciliakas: Changed them to stay on for 4 hours

PR #47, opened 2020-12-02
General improvements for mobile devices

Stokalas:
- Changed filters in all pages to be mobile responsive.
 - Footer always stays at the bottom and never floats. (NOT COMPATIBLE WITH CHROME 27; IE 8;  OPERA 12; SAFARI 8.0)
 - Validation error when adding goal manually now displayed on separate lines in Modal.
 - Reworked to fit mobile devices: addition of income and expenses.
 - Switched order of charts in dropdown to fit `Finances` dropdown.
 - Removed `@media` code from `CSS` as much as possible and used Bootstraps's grid system. (`col-3; col-md-6` etc.)
 - Mobile responsive expenses table? It is part why I'm creating this pull request without doing this for income. Since only way to solve wide tables on mobile is to make it side-scrollable (as far as I thought about it) and that  is not the best solution. I want you to test it and write here ideas how it could be changed/improved? Since simply making it smaller doesn't make sense (unreadable).

_I've tested everything on Google Chrome. As regular desktop website and as F12 mobile mode on iPhone X and iPad._

PS don't pay attention to the delete button on Expenses table, it was move outside table in preparations to make rows selectable and delete button above table. For now it was placed wherever.

### DIFF
--- a/src/Pages/Register.cshtml
+++ b/src/Pages/Register.cshtml
@@ -2,7 +2,7 @@
 @model ePiggyWeb.Pages.RegisterModel
 
 
-<div class="container login-container col-lg-6 col-md-8 col-sm-10 col-12">
+<div class="container login-container col-lg-4 col-md-8 col-sm-10 col-12">
     <div class="login-form-2">
         <h3>Sign Up</h3>
         <form autocomplete="off" method="post">
--- a/src/Pages/Register.cshtml
+++ b/src/Pages/Register.cshtml
@@ -3,7 +3,7 @@
 
 
 <div class="container login-container col-lg-4 col-md-8 col-sm-10 col-12">
-    <div class="login-form-2">
+    <div class="login-form-2 bg-primary">
         <h3>Sign Up</h3>
         <form autocomplete="off" method="post">
             <div asp-validation-summary="ModelOnly" class="text-danger"></div>
--- a/src/Pages/Register.cshtml
+++ b/src/Pages/Register.cshtml
@@ -2,7 +2,7 @@
 @model ePiggyWeb.Pages.RegisterModel
 
 
-<div class="container login-container">
+<div class="container login-container col-lg-6 col-md-8 col-sm-10 col-12">
     <div class="login-form-2">
         <h3>Sign Up</h3>
         <form autocomplete="off" method="post">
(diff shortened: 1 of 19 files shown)


src/Pages/Register.cshtml
```diff
@@ -2,7 +2,7 @@
 @model ePiggyWeb.Pages.RegisterModel
 
 
-<div class="container login-container">
+<div class="container login-container col-lg-6 col-md-8 col-sm-10 col-12">
```
ciliakas: I think 6 columns are a bit too large for desktop, I tried it out with 4 and it looks a bit more proportional in my opinion :)

ciliakas (commented):
Tested out both mobile and desktop versions, they work very well in my opinion.
About the tables, I think your current idea is quite good, because if you have mobile device horizontally, then the tables fit very well, it's just a bit awkward when the screen is vertical. 
My idea ( it might be difficult to implement, I'm not sure) is to have a drop down view (or like expanded view, like let's say it the entries only display name and price, but if you tap on it it expands and show the rest of the info about the entry) for the entries in the table

Stokalas:
I've thought about implementing drop down columns as was suggested, but since displaying different layouts according to the device is outdated, and better solution is to make responsive website, I've scrapped that idea since it would be weird for desktop users.
I've created two responsive designs - one in expenses page (reworked table) and one in income page (cards, similar to goals).
Both designs are just to get feel how it would look and may have some bugs since not at final version.

I would like to hear your opinion on what to do: which design is better? or maybe expenses should use one and income the other? I'm open for all ideas. Also report all bugs you'll come across 🙂.

--------------------------
I'm adding screenshots as well:
 - iPhone X expenses: 
image
 - iPad expenses: 
image
 - Desktop expenses: 
image
--------------------------
 - iPhone X income: 
image
 - iPad income: 
image
 - Desktop income:
image

Stokalas:
Ofcourse, cards could be wider and look like "chopped table". Example: 
image

PS Delete button in expenses table won't stay, I just needed it for testing purposes since delete selected is not working, and won't be included in final design.

ciliakas (commented):
Wow, good work! Love the fact that we get to two designs to choose from :D The reworked table design in expenses is the one I like more, in my opinion table design works better for entries than cards. Both look really good in either case!

DeividasLT5:
Didn't spot any bugs, as for the designs, both look good, have a hard time choosing, but I have to agree with Arnas that the expenses one is better in this context. Really good work, my only suggestions would be to think whether we should or shouldn't highlight the amount and importance somehow in expenses (or make the other data less visible?) as those other two don't really share the same importance to the user to see at all times. Also importance in income seems like a dead concept, so maybe just set it to a default state and not make the user choose it at all? Not sure if this is the right place to bring them up, but just things to think about.

MatasLa (approved):
Excellent improvements. Tested and didn't noticed any bugs. I agree with Arnas about design, tables looks much better for entries than cards. Also agree with Deividas about importance on income. I believe we should set it to default state.

PR #48, opened 2020-12-05
Multiple entries deletion

Stokalas:
_Small update that includes:_
 - Ability to delete selected multiple items in Expenses Table.
 - Mobile responsive tables for displaying "calculation's" result.

_Other small improvements:_
 - Improved expenses table design by removing some border lines, only one column in header, bigger checkbox (touchscreen users), edit button uses icon instead of text.
 - Added buttons inside index page when logged in for quicker navigation, especially for mobile users.
 - Fixed top padding in ChangePassword (account settings) page.
 - Changed Login/Register forms' color to better match theme.

_Update 2020-12-06:_
 - Ability to select/deselect all entries in expenses table.
 - Expenses table now has margins to the sides.
 - A lot of icons in the buttons. 
 - When changed filter in Suggestions crashing bug fix.
I've noticed that the table can still not fit on some very tight phone screens (such as Galaxy Fold), so in such cases I made the table horizontically scrollable.

### DIFF
--- a/src/DataManagement/MonthlyReport/MonthlyReportCalculator.cs
+++ b/src/DataManagement/MonthlyReport/MonthlyReportCalculator.cs
@@ -46,14 +46,19 @@ namespace ePiggyWeb.DataManagement.MonthlyReport
             var month = new DateTime(today.Year, today.Month, 1);
             var first = month.AddMonths(-1);
             var last = month.AddDays(-1);
-            var expenses = (await EntryDatabase.ReadListAsync(UserId, EntryType.Expense)).GetSum();
-            var income = (await EntryDatabase.ReadListAsync(UserId, EntryType.Income)).GetSum();
-            AllSavings = income - expenses;
-            Expenses = (await EntryDatabase.ReadListAsync(UserId, EntryType.Expense)).GetFrom(first).GetTo(last);
-            Income = (await EntryDatabase.ReadListAsync(UserId, EntryType.Income)).GetFrom(first).GetTo(last);
+
+            Expenses = await EntryDatabase.ReadListAsync(UserId, EntryType.Expense);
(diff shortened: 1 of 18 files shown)


ciliakas (commented):
Nice work! Having tested this out, I have a few suggestions:
- On Desktop, the table in expenses might look nicer with some padding on the sides
- Select all checkbox would be a great quality of life improvement :D 
- A little error I found is that the name of the table shifts to the right if you have an entry with a long name, like so:
bug1
bug2

Stokalas:
Was the long entry entered now? Or is it from the past, since now it should limit title to 25 or smth char length.

ciliakas:
Good catch, it's actually just an old entry :D

ciliakas (approved):
Good update, addressed the issues raised and more! The page is really coming more alive with the nicer buttons. 
Also as Deividas mentioned, going forward maybe we should remove the importance section from Incomes, at least in the front end, as in the end it doesn't seem we will be using it anywhere ( Of course it shouldn't be done in this pull request, just a thought as we move forwards )

Stokalas:
> 
> 
> Good update, addressed the issues raised and more! The page is really coming more alive with the nicer buttons.
> Also as Deividas mentioned, going forward maybe we should remove the importance section from Incomes, at least in the front end, as in the end it doesn't seem we will be using it anywhere ( Of course it shouldn't be done in this pull request, just a thought as we move forwards )

I agree with this and I think I'll rework Income next: display entries in tables like expenses, remove importance etc., just one question: if we don't use importance, then current chart for income is redundant, to what can we change it? I need ideas. 😆

MatasLa (approved):
Great job, like always. Nice improvements on design and ability to select multiple entries should ease of the use of ePiggy.

DeividasLT5 (approved):
Commented above instead of reviewed, whoops

PR #50, opened 2020-12-07
Calculation fix

DeividasLT5:
Mainly empty list / impossible to save because of data fixes (kind of WIP)

### DIFF
--- a/src/DataManagement/Saving/AlternativeSavingCalculator.cs
+++ b/src/DataManagement/Saving/AlternativeSavingCalculator.cs
@@ -19,7 +19,7 @@ namespace ePiggyWeb.DataManagement.Saving
             
             if (entryList.Count == 0)
             {
-                return new CalculationResults(entrySuggestions, monthlySuggestions, 0);              
+                return new CalculationResults(entrySuggestions, monthlySuggestions, -1);              
             }
 
             var enumCount = Enum.GetValues(typeof(Importance)).Length;
@@ -67,9 +67,16 @@ namespace ePiggyWeb.DataManagement.Saving
                 }
             }
             var timesToRepeatSaving = 0;
+            var firstTimeThroughWhile = true;
             var approximateSavedAmount = startingBalance;
             while (goal.Amount > approximateSavedAmount)
             {
+                if(!firstTimeThroughWhile && approximateSavedAmount <= startingBalance) //Can't possibly save for goal
+                {
+                    //inform user that his data is insufficient/show example lists
+                    return new CalculationResults(entrySuggestions, monthlySuggestions, -2);
+                }
+                firstTimeThroughWhile = false;
                 for (var i = enumCount; i > (int)Importance.Necessary; i--)
                 {
                     if (entryAmounts[i - 1] != 0)
--- a/src/DataManagement/Saving/AlternativeSavingCalculator.cs
+++ b/src/DataManagement/Saving/AlternativeSavingCalculator.cs
@@ -3,12 +3,11 @@ using System.Collections.Generic;
 using ePiggyWeb.DataManagement.Entries;
 using ePiggyWeb.DataManagement.Goals;
 using ePiggyWeb.Utilities;
-using Microsoft.AspNetCore.Mvc.RazorPages;
 using Microsoft.Extensions.Configuration;
 
 namespace ePiggyWeb.DataManagement.Saving
 {
-    public class AlternativeSavingCalculator : PageModel
+    public class AlternativeSavingCalculator
     {
         private static decimal RegularSavingValue { get; } = 0.25M;
         private static decimal MaximalSavingValue { get; } = 0.1M;
@@ -37,14 +36,7 @@ namespace ePiggyWeb.DataManagement.Saving
             for (var i = enumCount; i > (int)Importance.Necessary; i--)
             {
                 IEntryList expenses;
-                if (!generateRandomData)
-                {
-                    expenses = entryList.GetBy((Importance)i);
-                }
-                else
-                {
-                    expenses = expensesRandomList.GetBy((Importance)i);
-                }
+                expenses = !generateRandomData ? entryList.GetBy((Importance)i) : expensesRandomList.GetBy((Importance)i);
 
                 var ratio = enumCount - i;
                 foreach (var entry in expenses)
@@ -70,7 +62,7 @@ namespace ePiggyWeb.DataManagement.Saving
                             break;
                         default:
                             throw new Exception("Unexpected saving type");
-                    };
+                    }
                     entrySuggestions.Add(new SavingSuggestion(entry, amountAfterSaving));
 
                     sumsOfAmountByImportanceAdjusted[i - 1] += amountAfterSaving;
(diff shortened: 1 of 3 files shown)


src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -3,24 +3,23 @@
 using ePiggyWeb.DataManagement.Entries;
 using ePiggyWeb.DataManagement.Goals;
 using ePiggyWeb.Utilities;
+using Microsoft.AspNetCore.Mvc.RazorPages;
+using Microsoft.Extensions.Configuration;
 
 namespace ePiggyWeb.DataManagement.Saving
 {
-    public class AlternativeSavingCalculator
+    public class AlternativeSavingCalculator : PageModel
```
ciliakas: This class doesn't need to inherit from PageModel ( PageModel is only used in the front end files)

src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -30,9 +29,23 @@ public CalculationResults GetSuggestedExpensesOffers(IEntryList entryList, IGoal
...
-                var expenses = entryList.GetBy((Importance)i);
+                IEntryList expenses;
+                if (!generateRandomData)
+                {
+                    expenses = entryList.GetBy((Importance)i);
+                }
+                else
+                {
+                    expenses = expensesRandomList.GetBy((Importance)i);
+                }
```
ciliakas: This if else can be changed to '?' statement:

         expenses = !generateRandomData ? entryList.GetBy((Importance)i) : expensesRandomList.GetBy((Importance)i);

src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -92,7 +112,15 @@ public CalculationResults GetSuggestedExpensesOffers(IEntryList entryList, IGoal
...
             }
-            return new CalculationResults(entrySuggestions, monthlySuggestions, timesToRepeatSaving);
+            if (!generateRandomData)
+            {
+                return new CalculationResults(entrySuggestions, monthlySuggestions, timesToRepeatSaving);
+            }
+            else
+            {
+                return new CalculationResults(entrySuggestions, monthlySuggestions, timesToRepeatSaving * -1);
+            }
```
ciliakas: This if else also can be shortened:
       
     return !generateRandomData ? new CalculationResults(entrySuggestions, monthlySuggestions,mesToRepeatSaving) 
                : new CalculationResults(entrySuggestions, monthlySuggestions, timesToRepeatSaving * -1);

src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -57,7 +70,6 @@ public CalculationResults GetSuggestedExpensesOffers(IEntryList entryList, IGoal
                             break;
                         default:
                             throw new Exception("Unexpected saving type");
-                            break;
                     };
```
ciliakas: Unnecessary semicolon here

ciliakas (changes requested):
Good work so far, seems to be working a lot more stable now, and I don't think I was able to crash it even once now :D However I did leave a few comments on the code itself, and I was able to find what I think were glitches

bug21
bug22
 
For some reason, in this example, medium and maximal saving were showing the entries, but minimal was showing nothing ( I was also able to replicate this with example data (as in a date with no entries), I tried a few dozen times and every 5 or 10 times it happens)

Another thing is for the front end, I think it hasn't yet been updated to check if the returned int from calculations is negative or positive to check if it's example data or not

src/DataManagement/Saving/AlternativeSavingCalculator.cs
DeividasLT5: ah, okay, that makes more sense

src/DataManagement/Saving/AlternativeSavingCalculator.cs
DeividasLT5: Arnas, for that minimal not showing everything, it's a feature, it purposefully returns an empty list (a little bit of front-end notification is missing here). Because minimal saving only suggests you to lower "unnecessary" and "low" importances, i believe, so in a case where you only spent money on, for example, "high" importance expenses, theres no way to suggest you to save more in a "minimal" way.

DeividasLT5:
Arnas, for that minimal not showing everything, it's a feature, it purposefully returns an empty list (a little bit of front-end notification is missing here). Because minimal saving only suggests you to lower "unnecessary" and "low" importances, i believe, so in a case where you only spent money on, for example, "high" importance expenses, theres no way to suggest you to save more in a "minimal" way.

src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -37,14 +36,7 @@ public CalculationResults GetSuggestedExpensesOffers(IEntryList entryList, IGoal
...
                 IEntryList expenses;
-                if (!generateRandomData)
-                {
-                    expenses = entryList.GetBy((Importance)i);
-                }
-                else
-                {
-                    expenses = expensesRandomList.GetBy((Importance)i);
-                }
+                expenses = !generateRandomData ? entryList.GetBy((Importance)i) : expensesRandomList.GetBy((Importance)i);
```
ciliakas: And further on you can do this 


                var expenses = !generateRandomData ? entryList.GetBy((Importance)i) : expensesRandomList.GetBy((Importance)i);

src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -30,9 +28,16 @@ public CalculationResults GetSuggestedExpensesOffers(IEntryList entryList, IGoal
...
+
             for (var i = enumCount; i > (int)Importance.Necessary; i--)
             {
-                var expenses = entryList.GetBy((Importance)i);
+                IEntryList expenses;
+                expenses = !generateRandomData ? entryList.GetBy((Importance)i) : expensesRandomList.GetBy((Importance)i);
+
                 var ratio = enumCount - i;
                 foreach (var entry in expenses)
                 {
```
ciliakas, line 42: Can't comment on line 44 itself, but a way to simplify this would be:

       decimal amountAfterSaving;

(In this situation you don't need to assign a value with the declaration because you are overriding this value anyways in the code below it)

ciliakas (approved):
All good now, left only a couple of minor notices where I see things could be improved even further

src/DataManagement/Saving/AlternativeSavingCalculator.cs
DeividasLT5: hmm, but then we make the "var" into a "decimal", even if it does get rid of an unnecessary assignment, so why, exactly, is this better? The var is useless/unnecessary here?

ciliakas: Well if I understand correctly, you want to avoid unnecessary value assignment as happens in this case ( basically if you check your code below, you could assign  
      
      var amountAfterSaving = 999999999M

and it would not change how your code would work, so that's why I believe var should not be used in this case, even though we usually declare everything with var ( and to finish off Resharper is suggesting this change as well :D )

Stokalas: you should use var when you're assigning something to newly created variable, since in this case you don't have what to assign, yet, it's better to use `decimal _name_;` (atleast in my opinion)

DeividasLT5: oooh, ok, makes sense, i assumed incorrectly in a couple of ways :D

PR #51, opened 2020-12-07
Db requirements

MatasLa:
Usage of DataAdapter and DataTable in EntryDatabase.cs


src/DataBase/EntryDatabase.cs
```diff
@@ -178,7 +181,45 @@ public async Task<IEntry> ReadAsync(Expression<Func<IEntryModel, bool>> filter,
 
         public async Task<IEntryList> ReadListAsync(int userId, EntryType entryType)
         {
-            return await ReadListAsync(x => true, userId, entryType);
+            //return await ReadListAsync(x => true, userId, entryType);
+            var list = new EntryList(entryType);
+
+            var type = entryType == 0 ? "Incomes" : "Expenses";
+
+            string query = "SELECT * FROM " + type + " WHERE UserId =" + userId;
```
ciliakas: This and quite a few other variables afterwards in this method could be declared with var

src/DataBase/EntryDatabase.cs
```diff
@@ -178,7 +181,45 @@ public async Task<IEntry> ReadAsync(Expression<Func<IEntryModel, bool>> filter,
...
+            try
+            {
+                System.Data.SqlClient.SqlDataAdapter MyDataAdapter = new System.Data.SqlClient.SqlDataAdapter(query, CnStr);
+
+                MyDataAdapter.Fill(MyDataSet);
+            }
+            catch (Exception ex)
+            {
+                return null;
+            }
```
ciliakas: You don't have to catch exceptions here, we have exception handling everywhere in the website, so if anything goes wrong in this code, it will be caught and logged

src/DataBase/EntryDatabase.cs
```diff
@@ -178,7 +181,45 @@ public async Task<IEntry> ReadAsync(Expression<Func<IEntryModel, bool>> filter,
...
+                    Amount = row.Field<decimal>("Amount"),
+                    Date = row.Field<DateTime>("Date"),
+                    Importance = row.Field<int>("Importance"),
+                    Recurring = row.Field<bool>("IsMonthly"),
+                    Title = row.Field<string>("Title"),
+                    UserId = row.Field<int>("UserId")
+                } as IEntry;
+
+                items.Add(obj);
+            }
```
ciliakas: This could be replaced with linq, and I think it would help fulfill part of the 4 requirement:

            var items = (from DataRow row in table.Rows
            select new Entry
            {
                Id = row.Field<int>("Id"),
                Amount = row.Field<decimal>("Amount"),
                Date = row.Field<DateTime>("Date"),
                Importance = row.Field<int>("Importance"),
                Recurring = row.Field<bool>("IsMonthly"),
                Title = row.Field<string>("Title"),
                UserId = row.Field<int>("UserId")
            } as IEntry).ToList();

ciliakas (changes requested):
Seems to work well, just a left a few notes in the comments on thing I think could be possibly improved

src/DataBase/EntryDatabase.cs
```diff
@@ -14,9 +15,11 @@ namespace ePiggyWeb.DataBase
     public class EntryDatabase
     {
         private PiggyDbContext Database { get; }
+        public string CnStr { get; set; }
         public EntryDatabase(PiggyDbContext database)
         {
             Database = database;
+            CnStr = "Server=51.75.187.147;Database=SmartSaver;User Id=usern;Password=123456789;";
```
Stokalas, line 22: As I understand this is connection string, and I'm not sure but, we're saving it in app configs, so declaring it as string is a bad practice? @ciliakas could elaborate?

Stokalas (changes requested):
Seems to be working and fine, and most issues already has been fixed, just this one detail which I think should be taken into consideration.

src/DataBase/EntryDatabase.cs
ciliakas: Well I was thinking about the same thing, buuuut I think that since we're only using Data Adapter and Data Table for requirement purposes, and not actually what we would use in our app, I think this can slide. Anyway to be on the safe side we should ask the lecturer what he thinks

Stokalas (approved):

ciliakas (approved):

PR #52, opened 2020-12-08
Calculation fix display

Stokalas:
Did not test. Tried to do as fastly as possible.


ciliakas (approved):
Did some testing, seems to work fine

Stokalas:
Since this is really a small change, and nothing much to review, merging without waiting for any other confirmations.

PR #53, opened 2020-12-09
Currency Converter API

ciliakas:
This might look like a big pull request, but most of the files is just the set up logic of the API, and the whole business logic is in  CurrencyService, the routing is done by CurrencyController, errors are handled by ErrorHandlingMiddleWare.

In short, this API has 4 get routes:
 - /currency - get information about a currency (it's full name, it's rate against Euro, and it's symbols as an array of unicode decimals) by it's code (EUR, USD, GBP etc.)
 - /list - get info about all currencies that we keep track of the rate for
 - /rate - get specific rate between two currencies 
 - /symbols - get all the symbol info that we store, _without rate info_

You can see an example of how it works in the Entry.cshtml.cs file.

The rates are taken from the European Central Bank.


.gitignore
```diff
@@ -32,6 +32,7 @@ bld/
 
 # Logs
 Logs/
+log*
```
Stokalas, line 35: This should ignore all files that has log at the start of name? (never really paid attention to how .gitignore works :D )

CurrencyConverter/CurrencyConverter.Domain/Currency.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace CurrencyConverter.Domain
+{
+    public class Currency
```
Stokalas, line 6: What's the difference between this and CurrencyDto? Or since they look the same, where one is used and there another?

CurrencyConverter/CurrencyConverter.WebApi/Middleware/ErrorHandling/ErrorHandlingMiddleware.cs
```diff
@@ -0,0 +1,76 @@
...
+                await HandleExceptionAsync(context, ex);
+            }
+        }
+
+        private Task HandleExceptionAsync(HttpContext context, Exception exception)
+        {
+            context.Response.ContentType = "application/json";
+            context.Response.StatusCode = (int)HttpStatusCode.InternalServerError;
+
+            var errorMessage = "Internal server error happened. Please contact support";
```
Stokalas, line 46: Where is this outputted? Since as I understand the exception with it's stuff is logged in log file insider server?

CurrencyConverter/CurrencyConverter.Services/Services/CurrencyService.cs
```diff
@@ -0,0 +1,112 @@
...
+                    Code = node.Attributes["currency"].Value,
+                    Rate = decimal.Parse(node.Attributes["rate"].Value, NumberStyles.Any, new CultureInfo("en-Us"))
+                };
+                list.Add(rate);
+            }
+
+            return list.ToArray();
+        }
+
+        public IEnumerable<Currency> GetCurrencyList()
```
Stokalas, line 46: This basically calls methods that return rates from ECB and then symbols, and "joins" them into single list?

CurrencyConverter/CurrencyConverter.Services/Services/CurrencyService.cs
```diff
@@ -0,0 +1,112 @@
...
+            var currencies = list as Currency[] ?? list.ToArray();
+            var currency1 = currencies.FirstOrDefault(x => x.Code == currencyCode1);
+            var currency2 = currencies.FirstOrDefault(x => x.Code == currencyCode2);
+
+            if (currency1 is null || currency2 is null)
+            {
+                return null;
+            }
+
+            var rate = decimal.Round(currency2.Rate / currency1.Rate, 4);
```
Stokalas, line 72: Stand point for rates is EUR if I'm not mistaken?

Stokalas (approved):
Did not test, since I'm keen on testing when we will connect this with front, and now I trust that it works just fine. Did not notice anything wrong in code. Really well done! Just as always, few questions 🙂 
Approving

.gitignore
ciliakas: As I understand, yes :D And it seems to be working as intended

CurrencyConverter/CurrencyConverter.Domain/Currency.cs
ciliakas: Short answer, I'm not sure exactly why this is needed, but it has something to do with how API's should be designed and for API to be more reusable and loosely coupled.

https://en.wikipedia.org/wiki/Data_transfer_object

CurrencyConverter/CurrencyConverter.WebApi/Middleware/ErrorHandling/ErrorHandlingMiddleware.cs
ciliakas: It's logged to the file, and also the user of API would get this message as a response as well ( and if it's in development mode, the error code with it as well)

CurrencyConverter/CurrencyConverter.Services/Services/CurrencyService.cs
ciliakas: Yup, that's exactly what it does

CurrencyConverter/CurrencyConverter.Services/Services/CurrencyService.cs
ciliakas: I'm not sure I understand the question :D

Stokalas: For example, each currency has `.Rate`, and that rate is from the view (can't find the word:D) of EUR?
aka The EUR is the base currency for all the rates?

ciliakas: Yeah, EUR is the base currency for all rates, but if you want to get a specific rate between two currencies than you can use this method :D

PR #54, opened 2020-12-09
Mobile income

Stokalas:
- _Income_ is displayed in table, just like _Expenses_.
- Changed _Importance_ to _Attainability_ in _Income_.
- Changes in _Edit Entry_ and _Income chart_ pages to fit _Attainability_.
- Small changes in _Edit Entry_ to make it more "mobile friendly".
- Skeleton form in _Account settings_ to change currency.

### DIFF
--- a/src/Pages/ChangePassword.cshtml
+++ b/src/Pages/ChangePassword.cshtml
@@ -53,7 +53,7 @@
             <input class="btn btn-outline-warning" type="submit" value="Confirm" />
             <p class="text-warning">@Model.ErrorMessage</p>
             <hr />
-            <button type="button" class="btn btn-danger" data-toggle="modal" data-target="#ConfirmModal">Delete Account<i class="ml-1 fas fa-user-times"></i></button>
+            <button type="button" class="btn btn-sm btn-danger" data-toggle="modal" data-target="#ConfirmModal">Delete Account<i class="ml-1 fas fa-user-times"></i></button>
         </form>
     </div>
 </div>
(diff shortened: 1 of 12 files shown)


src/Pages/ChangePassword.cshtml
```diff
@@ -16,8 +16,28 @@
...
+        <hr/>
+        <h5 class="text-light">Currency</h5>
+        <form method="post">
+            <div class="row justify-content-center">
+                <select class="mb-3" asp-for="Currency">
+                    <option hidden disabled selected value="">-Select-</option>
+                    <option value="1">EUR</option>
+                    <option value="2">USD</option>
+                    <option value="3">GBP</option>
+                    <option value="4">RUB</option>
```
ciliakas: Could you make it possible for these values to be added from the cshtml.cs file itself? Like in the OnGet method we could set some values to a List and those would be displayed?

ciliakas (commented):
Everything looks good, just left a question in the comments

src/Pages/ChangePassword.cshtml
Stokalas: Yes, this is placeholder, and it is possible to do `foreach` which generates these "options" for each item in the thing we are iterating. (as I understand in this case List)

But since we're not assigning anything to list for now (API not implemented) it would not compile/render page, so placeholder was used, to check how it would look.

ciliakas: The thing is I'd like to start to try and implement the API, so I was wondering maybe you could do that foreach, just hardcode the list in the cshtml.cs, as in:

       var list = new List { "EUR", "USD", "GBP", "RUB"};
       foreach(var currency in list)
       { ... //set options ...}

Basically I want to have the base for me to easier implement the API :D

Stokalas: oh ok, will be done soon, maybe even today :D

ciliakas (approved):
With the changes I requested I was now able to start to implement the API :D

PR #56, opened 2020-12-11
Pagination

Stokalas:
Basic pagination to test in Expenses, if it works fine and after some code refactoring same will be added to income page. 
I've commented a part where backend should take a look.

### DIFF
--- a/src/DataManagement/Entries/Entry.cs
+++ b/src/DataManagement/Entries/Entry.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.ComponentModel.DataAnnotations;
 using System.Text;
 using ePiggyWeb.DataBase.Models;
 using ePiggyWeb.DataManagement.Goals;
@@ -10,7 +11,9 @@ namespace ePiggyWeb.DataManagement.Entries
     {
         public int Id { get; set; }
         public int UserId { get; set; }
+        [StringLength(25, ErrorMessage = "Too long title!")]
         public string Title {get; set; }
+        [Range(0, 99999999.99, ErrorMessage = "Amount out of range!")]
         public decimal Amount { get; set; }
         public DateTime Date { get; set; }
         public bool Recurring { get; set; }
--- a/src/Pages/Expenses.cshtml.cs
+++ b/src/Pages/Expenses.cshtml.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
-using System.Diagnostics;
 using System.Linq;
 using System.Security.Claims;
 using System.Threading.Tasks;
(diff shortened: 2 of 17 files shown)


src/DataManagement/Entries/Entry.cs
```diff
@@ -10,7 +11,9 @@ public class Entry : IEntry
     {
         public int Id { get; set; }
         public int UserId { get; set; }
+        [StringLength(25, ErrorMessage = "Too long title!")]
```
ciliakas, line 14: What are these attributes for?

src/Pages/Expenses.cshtml.cs
```diff
@@ -52,6 +53,13 @@ public class ExpensesModel : PageModel
         private EntryDatabase EntryDatabase { get; }
         private IConfiguration Configuration { get; }
 
+        [BindProperty(SupportsGet = true)]
+        public int CurrentPage { get; set; } = 1;  
+        public int PageSize = 10;
```
ciliakas, line 58: This could be converted to a constant

src/Pages/Expenses.cshtml.cs
```diff
@@ -143,14 +151,16 @@ private async Task SetData()
             {
                 UserId = int.Parse(User.FindFirst(ClaimTypes.Name).Value);
                 var entryList = await EntryDatabase.ReadListAsync(UserId, EntryType.Expense);
-                Expenses = entryList.GetFrom(StartDate).GetTo(EndDate);
-                AllExpenses = entryList.GetSum();
+                Expenses = entryList.GetFrom(StartDate).GetTo(EndDate);//If we want to remove this I need to get total amount of these as well as their sum
+                ExpensesToDisplay = Expenses.OrderByDescending(x=> x.Date).Skip((CurrentPage - 1) * PageSize).Take(PageSize);//I think this should be done by backend?
+                AllExpenses = Expenses.GetSum();
```
ciliakas: I'll try to move as much as I can to the back end, but it won't be trivial to do so it will take some time. Otherwise, this works really well and I like how you implemented pagination

ciliakas (commented):
Amazing work, works really well and a needed feature. Just left a few questions in the comments, and also I noticed a couple of other things not related to this pull request.
First is on Expenses and Incomes, OnPostNewEntry methods do not implement the cookie filter saving ( because I forgot to change them :D)
Second is I think the old commented out delete method can be removed

src/DataManagement/Entries/Entry.cs
Stokalas: It should limit Length to 25, and when longer entered this message should be displayed. But as far as I tested it just blocks entering chars when length is 25 in the field, so error message is redundant. Why I put this inside here? Just to show overall limitation, because same is used for each entry.

PR #57, opened 2020-12-14
Back Update for Pagination

ciliakas:
Added pagination to back end and added casting to IEntryList

### DIFF
--- a/src/DataManagement/Entries/EntryList.cs
+++ b/src/DataManagement/Entries/EntryList.cs
@@ -17,13 +17,19 @@ namespace ePiggyWeb.DataManagement.Entries
             EntryType = entryType;
         }
 
+        public EntryList(IEnumerable<IEntry> entryList)
+        {
+            if (entryList is null) throw new ArgumentException();
+            AddRange(entryList);
+        }
+
         private EntryList(EntryType entryType, IEnumerable<IEntry> entryList)
         {
             EntryType = entryType;
-            if (entryList is null) return;
+            if (entryList is null) throw new ArgumentException();
             AddRange(entryList);
         }
-        
+
         public IEntryList GetBy(Importance importance)
         {
             return new EntryList(EntryType, this.Where(x => x.Importance == (int)importance).ToList());
@@ -52,6 +58,11 @@ namespace ePiggyWeb.DataManagement.Entries
         public IEntryList GetUntilToday()
         {
             return new EntryList(EntryType, this.Where(x => x.Date <= DateTime.Today).ToList());
+        }
+
+        public IEntryList GetPage(int pageNumber, int pageSize)
+        {
+            return new EntryList(EntryType, this.Skip((pageNumber - 1) * pageSize).Take(pageSize).ToList());
         }
 
         public DateTime GetOldestEntryDate()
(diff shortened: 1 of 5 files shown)


src/Utilities/ArrayExtension.cs
```diff
@@ -1,9 +1,16 @@
...
+using System.Collections.Generic;
+using ePiggyWeb.DataManagement.Entries;
 
 namespace ePiggyWeb.Utilities
 {
     public static class ArrayExtension
     {
+        public static IEntryList ToIEntryList(this IEnumerable<IEntry> entryEnumerable)
+        {
+            return new EntryList(entryEnumerable);
```
Stokalas, line 11: How does this overcome problem of not being able to cast it simply?

src/Pages/Expenses.cshtml.cs
```diff
@@ -152,7 +152,8 @@ private async Task SetData()
                 UserId = int.Parse(User.FindFirst(ClaimTypes.Name).Value);
                 var entryList = await EntryDatabase.ReadListAsync(UserId, EntryType.Expense);
                 Expenses = entryList.GetFrom(StartDate).GetTo(EndDate);//If we want to remove this I need to get total amount of these as well as their sum
-                ExpensesToDisplay = Expenses.OrderByDescending(x=> x.Date).Skip((CurrentPage - 1) * PageSize).Take(PageSize);//I think this should be done by backend?
+                ExpensesToDisplay = Expenses.OrderByDescending(x => x.Date).ToIEntryList()
+                    .GetPage(CurrentPage, PageSize);
```
Stokalas: A bit of inconsistency, since in income this is one line. Or is this just github stuff? :D

src/Pages/Income.cshtml.cs
```diff
@@ -140,7 +140,7 @@ private async Task SetData()
                 UserId = int.Parse(User.FindFirst(ClaimTypes.Name).Value);
                 var entryList = await EntryDatabase.ReadListAsync(UserId, EntryType.Income);
                 Income = entryList.GetFrom(StartDate).GetTo(EndDate);
-                IncomeToDisplay = Income.OrderByDescending(x => x.Date).Skip((CurrentPage - 1) * PageSize).Take(PageSize);
+                IncomeToDisplay = Income.OrderByDescending(x => x.Date).ToIEntryList().GetPage(CurrentPage, PageSize);
```
Stokalas, line 143: //This is multiline in expenses I was talking about

src/DataManagement/Entries/EntryList.cs
```diff
@@ -17,13 +17,19 @@ public EntryList(EntryType entryType)
             EntryType = entryType;
         }
 
+        public EntryList(IEnumerable<IEntry> entryList)
+        {
+            if (entryList is null) throw new ArgumentException();
+            AddRange(entryList);
```
Stokalas, line 23: Ofc I just could google, but what does this addrange do? Does it just adds everything it received as parameter into "itself"?

Stokalas (approved):
It's always nice to move some of the logic behind. Left a few questions. Just not sure about pagination being in the back. Since it is related to how data is displayed and kinda front thing. Would really like to ask lecturer about this :D But still nice

src/Utilities/ArrayExtension.cs
ciliakas: Well as I understand, casting collections of objects is different than casting objects, for example you usually see methods like "ToArray", "ToList", etc., so I instead of trying to figure out a convoluted cast, I just followed this simple pattern and it seems like a much neater and easier solution, at least to me :D

src/Pages/Expenses.cshtml.cs
ciliakas: I think one of these automatically moved to a double liner, I'll fix it :D

src/DataManagement/Entries/EntryList.cs
ciliakas: Since the class is an extended List, it's just the same as AddRange for any other list, so in this case it justs add everything into itself, yes :D

PR #58, opened 2020-12-14
API Implementation methods

ciliakas:
Added classes and methods to work with the API, and an example in Entry.cshtml.cs to see how it works.
The API must be running in order for this to work.


src/CurrencyAPI/Currency.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        public decimal Rate { get; set; }
+
+        public string GetSymbol()
+        {
+            var sb = new StringBuilder();
+
+            foreach (var symbol in Symbol)
+            {
+                sb.Append(char.ConvertFromUtf32(symbol));
```
Stokalas, line 22: Why is it stored in such format?

src/DataBase/EntryDatabase.cs
```diff
@@ -160,12 +160,6 @@ public async Task<bool> DeleteListAsync(IEnumerable<int> idArray, int userId, En
 
         public async Task<IEntry> ReadAsync(int id, int userId, EntryType entryType)
         {
-            IEntryModel dbEntry = entryType switch
-            {
-                EntryType.Income => await Database.Incomes.FirstOrDefaultAsync(
-                    x => x.Id == id && x.UserId == userId),
-                _ => await Database.Expenses.FirstOrDefaultAsync(x => x.Id == id && x.UserId == userId)
-            };
```
Stokalas, line 168: Why was this removed?

src/DataBase/UserDatabase.cs
```diff
@@ -132,5 +129,49 @@ public async Task<UserModel> GetUserAsync(Expression<Func<UserModel, bool>> filt
...
+        public async Task ChangeCurrency(int userId, string currencyCode)
+        {
+            var user = await Database.Users.FirstOrDefaultAsync(x => x.Id == userId);
+
+            //user.Currency = currencyCode;
+
+            await Database.SaveChangesAsync();
+        }
+
+        public async Task ChangeCurrency(int userId, string currencyCode, decimal rate)
```
Stokalas, line 142: This is the one which recalculates as well?

src/CurrencyAPI/CurrencyConverter.cs
```diff
@@ -0,0 +1,66 @@
...
+            var message = await SendRequest(ConnectionString + RateRoute + "?currencyCode1=" + currencyFrom + "&currencyCode2=" + currencyTo);
+            var rate = JsonConvert.DeserializeObject<decimal>(message);
+            return rate;
+        }
+
+        public async Task<IList<Currency>> GetSymbols()
+        {
+            var message = await SendRequest(ConnectionString + SymbolsRoute);
+            var listDto = JsonConvert.DeserializeObject<List<CurrencyDto>>(message);
+            IList<Currency> list = listDto.Select(dto => dto.FromDto()).ToList();
```
Stokalas, line 55: Why not `var` here?

Stokalas (approved):
Nice practical implementation. Left few questions as always :). Other than that code seems fine, and works as I've used some of this when creating display table for all currencies.

src/CurrencyAPI/Currency.cs
ciliakas: Well shortest answer is because that what I found on the internet, but I figure why the symbols are stored as an array of integers - it is because the currency symbols can contain unique symbols so it is easier to transfer them as their unicode representations, and an array because some currencies only use one symbol and some use more than one

src/DataBase/EntryDatabase.cs
ciliakas: This was unused code left by mistake actually, I only left the line used by the method

src/DataBase/UserDatabase.cs
ciliakas: It's actually the other one, maybe I should name them differently :D At the moment the one that takes a decimal rate argument is the one that changes the rate, and this one doesn't touch it

src/CurrencyAPI/CurrencyConverter.cs
ciliakas: I prefer to use IList instead of List  ( I read somewhere that it is better to use than a list, but I can't say for sure why that is exactly), and ToList method returns a List and not an IList, so I have to manually assign it as such

src/DataBase/UserDatabase.cs
Stokalas: Atleast it shows to me that this one is the one that takes rate :D so yeah this should recalculate :D

ciliakas: Ah I was doing this too late in the night, you're right, the one that is showing rate is this one and it's the one that recalculates :D

PR #59, opened 2020-12-14
Currency display

Stokalas:
- Added `Linq Aggregate` when searching for the cheapest and the most expensive goals.
 - Simple page that displays all currencies and rates.


src/DataManagement/MonthlyReport/MonthlyReportCalculator.cs
```diff
@@ -104,22 +104,34 @@ private void CalculateGoalReport()
...
-            var min = decimal.MaxValue;
-            var minGoal = new Goal();
-            var maxGoal = new Goal();
-            var max = 0M;
-            foreach (var item in Goals)
+           
+          
+
+            var maxGoal = Goals.Aggregate((expensive, next) => next.Amount >= expensive.Amount ? next : expensive);
+            var minGoal = Goals.Aggregate((cheapest, next) => (next.Amount <= cheapest.Amount && next.Amount >= AllSavings)  ? next : cheapest);
```
ciliakas, line 111: So how does aggregate work? Does it just go through the list and applies a certain function as it goes?

src/DataManagement/MonthlyReport/MonthlyReportCalculator.cs
```diff
@@ -104,22 +104,34 @@ private void CalculateGoalReport()
...
-                    max = item.Amount;
-                    maxGoal = (Goal)item;
-                }
-
-                if (item.Amount > min || item.Amount <= AllSavings) continue;
-                min = item.Amount;
-                minGoal = (Goal)item;
+                Result.HasGoals = false;
+                return;
             }
```
ciliakas, line 117: So with aggregate you were able to shorten this much code? Seems like a neat method to use

ciliakas (approved):
Left a few questions, however everything looks well implemented. We just need to think of a place where to actually put the button to see the currency conversion rates

src/DataManagement/MonthlyReport/MonthlyReportCalculator.cs
Stokalas: Main concept is that variables `expensive` and ` cheapest` are for keeping results, and next is the list item currently in iteration. And usage of ternary operator allows to assign values to these "result" variables according to the condition. And after iterating it returns what's in that "result" variable. There is  ability to modify result in the end(e.g. divide by the number of items to get average if we're adding items) but it was not needed here. At least that's how I understand it

src/DataManagement/MonthlyReport/MonthlyReportCalculator.cs
Stokalas: Basically yes, since it allows to avoid ifs and no need to have middle vars such as max and min (but it could be avoided if goals were comparable by IComparable interface by amounts (mayybe we have it but eh :D)).

PR #60, opened 2020-12-14
Added Currency to User, set value to 'EUR'

MatasLa:
Just added Currency to User and set value to EUR for existing users.


src/DataBase/Models/UserModel.cs
```diff
@@ -10,7 +10,8 @@ public class UserModel
         public string Email { get; set; }
         public string Password { get; set; }
         public string Salt { get; set; }
-
+        [StringLength(3)]
+        public string Currency { get; set; }
```
Stokalas, line 14: The question that I'm sure was answered but to make it sure: migration files are generated automatically and we should not pay attention to them?

Stokalas (approved):
If I understand it correctly migration files are auto-generated. So nothing much to look at here. But nice, since it will allow to implement the API

ciliakas (approved):
Everything works fine, just want to know the answers to what Gytis is asking as well :D

src/DataBase/Models/UserModel.cs
MatasLa: Migration file needed some changes, because there were errors. I had to edit migration, because we are using another DbContext now and initial migration were not valid, so I manually edited it to only add column. (auto generated migration file was creating all tables again)

PR #61, opened 2020-12-15
Db requirements

MatasLa:
Created class GoalDatabaseSql.cs similar to GoalDatabase.cs
Created methods with select, insert, delete, update
Method with DataAdapter and DataTable
Removed SQL method from EntryDatabase


ciliakas (commented):
Everything is going in the right direction - the only thing that is left to do is to fully implement all other methods and create an interface for both of the Database classes

src/DataBase/GoalDatabaseSql.cs
```diff
@@ -0,0 +1,203 @@
...
+            var sqlConnection = new SqlConnection(Database.Database.GetDbConnection().ConnectionString);
+
+            if (sqlConnection.State == ConnectionState.Closed)
+            {
+                sqlConnection.Open();
+            }
+
+            var dbGoal = new GoalModel(goal, userid);
+
+            var sqlCommand = new SqlCommand("INSERT INTO Goals(UserId, Price, Title) VALUES (@UserId, @Price, @Title);SELECT CAST(scope_identity() AS int);", sqlConnection);
```
DeividasLT5, line 36: This line doesn't even fit in github (on files changed screen for example), so probably you should chop it somehow?

DeividasLT5 (commented):
Seems good, however I have a small note unrelated to the review. Noticed how you deleted a couple commented code chunks, so I wanted to remind you that dead code can be taken from previous commits if needed. Just in case it was kept there a while for "maybe we'll need it" :)

PR #62, opened 2020-12-15
Currency API Integration

ciliakas:
Integrated the Currency API to display different currencies and convert data if needed. There's quite a few files changed, but most are just refactoring code to take in the new parameters to display currency. Main files to look at:
- CurrencyConverter - added a method to just get a symbol for the currency code given, because this is the most common use of the API
- ChangePassword - added the selection of currencies, set the default selected option to be the current currency of the user. If there is a problem with the API, no other currencies are displayed, only the one that user has. _Implementing an error message here would be good_
- NumberFormatter - added the optional parameter to display a different symbol for currency. Because the currency symbol parameter is optional, it doesn't brake any code anywhere else
- The rest of the file are pretty much analogous, we set the currency symbol and that is display everywhere where currency is displayed, however if the API fails it just displays the currency code instead _An error message here would be good_

Notes:
 - Improved the accuracy of conversion by rounding the rates, so if we want to display them properly we need to set how many digits to display manually.
 - To improve performance, we could store the symbols in the cookies, because at the moment every time a page is loaded it needs to go trough the API which affects performance


CurrencyConverter/CurrencyConverter.Services/Services/CurrencyService.cs
```diff
@@ -69,7 +69,7 @@ public IEnumerable<Currency> GetCurrencyList()
                 return null;
             }
 
-            var rate = decimal.Round(currency2.Rate / currency1.Rate, 4);
+            var rate = currency2.Rate / currency1.Rate;
```
DeividasLT5, line 72: Sorry, I'm confused "Improved the accuracy of conversion by rounding the rates, so if we want to display them properly we need to set how many digits to display manually." seems like you removed rounding? So was your note just phrased incorrectly, or what do you mean? I just see removal of rounding, and I can see how that would improve accuracy, but i don't see "Improved the accuracy of conversion by rounding the rates,"  :D

ciliakas: Yeah, I meant removing rounding of rates, :D

src/Pages/ChangePassword.cshtml.cs
```diff
@@ -58,15 +59,30 @@ public async Task<IActionResult> OnGet()
...
+                var currencyList = await CurrencyConverter.GetList();
+                foreach (var currency in currencyList)
+                {
+                    CurrencyOptions.Add(currency.Code);
+                }
+            }
+            catch (Exception)
+            {
+                CurrencyOptions.Add(UserModel.Currency);
+                FailedToGetCurrencyList = true;
```
Stokalas: As I understand most of code parts like these were changed when caching was implemented, so maybe it is not worth reviewing this?

Stokalas (approved):
Well this worked when tested, but now since most of code changed maybe time to merge these?

src/Pages/ChangePassword.cshtml.cs
ciliakas: Yeah, I've already updated them in caching, and now updated this code again in the new branch I'm working on

PR #63, opened 2020-12-16
Added alert when unable to read currencies

Stokalas:
This is a really small PR, I'm not sure if this is all, but I've found only one comment where alert should be. Inform me if I missed some.

### DIFF
--- a/src/Pages/ChangePassword.cshtml
+++ b/src/Pages/ChangePassword.cshtml
@@ -19,6 +19,12 @@
         }
         <h5 class="text-light mb-3 mt-3">User: <span class="text-warning">@User.FindFirst(ClaimTypes.Email).Value</span></h5>
         <hr />
+        @if (Model.FailedToGetCurrencyList)
+        {
+            <div class="alert alert-danger" role="alert">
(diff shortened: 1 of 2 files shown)


ciliakas:
This is exactly what was needed, but I forgot to explain properly where else these should be implemented :D  Basically, this alert is needed everywhere, where we display currency : Incomes, Expenses, Goals, Suggestions, Monthly Reports, all the graphs.
I think the only place where we don't need this alert is in EditEntry, where we don't display any currency signs

Stokalas:
But as I understand when it does not load it displays the EUR instead of eur symbol, so is there really a need to bother user with stuff like that? Not that it would be hard to implement these alerts, but in Account settings alert is needed to inform that it is not possible to change the currency, but where the entries are displayed it does not stop the main flow of those pages, and the alert would be kinda "weird". At least that's how I see it :D

ciliakas:
That's a really good point, but now that we will be just converting currency on the front end when we will be displaying it, probably we will want to have this alert anyway, so that people know we can't display in their selected currency

PR #64, opened 2020-12-16
Small fixes

Stokalas:
- Tried to enforce currency dropdown in Account settings to always open downwards by changing layout.
- New Tab data in navbar which is new home for currency rates table
- When failed to parse goal alert displayed.

PS. This is from dev so in order to open Account settings you need to have currency API running.
PSS. To try to fail goal parse I entered 'asdasdasdasdasdasdasdasdasd' or smth like that :)

### DIFF
--- a/src/Pages/ChangePassword.cshtml
+++ b/src/Pages/ChangePassword.cshtml
@@ -2,12 +2,12 @@
 @using System.Security.Claims
 @model ePiggyWeb.Pages.ChangePasswordModel
 
-<div class="container login-container col-lg-4 col-md-8 col-sm-10 col-11 mt-5">
+<div class="container login-container col-lg-4 col-md-8 col-sm-10 col-11 mt-2">
     <div class="login-form-2 bg-primary border">
-        <h5 class="text-light mb-3 mt-3">User: <span class="text-warning">@User.FindFirst(ClaimTypes.Email).Value</span></h5>
-        <hr/>
+        <h5 class="text-light mt-3">User: <span class="text-warning">@User.FindFirst(ClaimTypes.Email).Value</span></h5>
+        <hr />
         <h5 class="text-light">Currency</h5>
-        <form method="post">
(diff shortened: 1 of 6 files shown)


ciliakas (changes requested):
The drop down now looks better for me - doesn't open upwards anymore, so even if it quite large I think this is plenty good enough for our needs.
I like the location for the currency data, I think we also need to add alerts there if the API is down ( however this might make more sense to do on the other pull request :D)
No matter what I tried I couldn't get the parser to throw me an exception - anything I wrote still somehow got a price, my favorite attempt was "wfcdv 34t 1e6rg z", which costs 14.11 :D 

The only reason why I'm requesting changes, is that for some reason in account settings when I click on snow or change password button, I get the little drop down with my browser's saved passwords and nothing else happens, so I can't turn on snow or change password :D

Stokalas:
You can always check the page for what was parsed, this time [it was](https://uk.camelcamelcamel.com/search?sq=wfcdv+34t+1e6rg+z) , so weird but works :D I will take a look at the problem tomorrow, thanks for reporting.

DeividasLT5:
I think we shouldn't allow things like "wfcdv 34t 1e6rg z" to get a price? Even in that specific case, I couldn't find the actual product named that, so it more or less "parsed the internet link"? :D Seems unnatural, no? Is this being worked on?

Stokalas:
> 
> 
> I think we shouldn't allow things like "wfcdv 34t 1e6rg z" to get a price? Even in that specific case, I couldn't find the actual product named that, so it more or less "parsed the internet link"? :D Seems unnatural, no? Is this being worked on?

We could do a regex to check for a valid name, but again some things e.g. car parts has special codes, so in some cases weird things could be valid and limiting user when things are being found on Camel is not necessary. And if item is not found then it would be displayed as we could not parse the data :)

ciliakas (approved):

PR #65, opened 2020-12-17
Currency Info Caching

ciliakas:
Only review this after looking through the previous API implementation pull request :D 
Implemented caching to give quite a big performance boost and efficiency. The only files to look at are:
 - Change Password: Added caching of the list of currencies, also caches user currency if not already cached
 - Expenses

All the others are just analogues of Expenses or utilities to implement the caching.

Notes:
 - I set the cache to last either 15 minutes sliding, to the (absolute) limit of 16:00 UTC time, because that is the time currency rates refresh
 - When caching, if I don't find the cache and the API doesn't work, I don't cache the backup values (currency codes etc.) that we display

### DIFF
--- a/src/Pages/ChangePassword.cshtml.cs
+++ b/src/Pages/ChangePassword.cshtml.cs
@@ -164,6 +164,9 @@ namespace ePiggyWeb.Pages
         public async Task<IActionResult> OnPostDeleteAccount()
         {
             await HttpContext.SignOutAsync();
+            Response.Cookies.Delete("StartDate");
+            Response.Cookies.Delete("EndDate");
+            Cache.Remove(CacheKeys.UserCurrency);
             await UserDatabase.DeleteUserAsync(User.FindFirst(ClaimTypes.Email).Value);
             return Redirect("/index");
         }
--- a/src/Pages/ChangePassword.cshtml.cs
+++ b/src/Pages/ChangePassword.cshtml.cs
@@ -76,6 +76,23 @@ namespace ePiggyWeb.Pages
             if (Cache.TryGetValue(CacheKeys.CurrencyList, out List<string> cachedCurrencyList))
             {
                 CurrencyOptions.AddRange(cachedCurrencyList);
+                //Check if user's currency is cached
+                if (Cache.TryGetValue(CacheKeys.UserCurrency, out Currency cachedCurrency))
+                {
+                    if (UserModel.Currency.Equals(cachedCurrency.Code)) return;
+                }
+
+                //If we the code gets here, it means that we don't have the the correct user currency cached
+                try
+                {
+                    var userCurrency = await CurrencyConverter.GetCurrency(UserModel.Currency);
+                    var options = CacheKeys.DefaultCurrencyCacheOptions();
+                    Cache.Set(CacheKeys.UserCurrency, userCurrency, options);
+                }
+                catch (Exception e)
+                {
+                    _logger.LogInformation(e.ToString());
+                }
                 return;
             }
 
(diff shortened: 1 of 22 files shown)


src/Pages/ChangePassword.cshtml.cs
```diff
@@ -43,14 +45,16 @@ public class ChangePasswordModel : PageModel
         private EmailSender EmailSender { get; }
         private CurrencyConverter CurrencyConverter { get; }
         public UserModel UserModel { get; private set; }
+        private IMemoryCache Cache { get; }
 
-        public ChangePasswordModel(PiggyDbContext piggyDbContext, IOptions<EmailSender> emailSenderSettings, ILogger<ChangePasswordModel> logger, CurrencyConverter currencyConverter)
+        public ChangePasswordModel(PiggyDbContext piggyDbContext, IOptions<EmailSender> emailSenderSettings, ILogger<ChangePasswordModel> logger, CurrencyConverter currencyConverter, IMemoryCache cache)
```
Stokalas, line 51: This line does not fit the screen on github, so I think it would be good to chop it.

src/Pages/Expenses.cshtml.cs
```diff
@@ -61,14 +62,17 @@ public class ExpensesModel : PageModel
         public bool ShowNext => CurrentPage < TotalPages;
         private UserDatabase UserDatabase { get; }
         private CurrencyConverter CurrencyConverter { get; }
+        public decimal CurrencyRate { get; set; }
         public string CurrencySymbol { get; private set; }
-        public ExpensesModel(EntryDatabase entryDatabase, ILogger<ExpensesModel> logger, IConfiguration configuration, UserDatabase userDatabase, CurrencyConverter currencyConverter)
+        private IMemoryCache Cache { get; }
+        public ExpensesModel(EntryDatabase entryDatabase, ILogger<ExpensesModel> logger, IConfiguration configuration, UserDatabase userDatabase, CurrencyConverter currencyConverter, IMemoryCache cache)
```
Stokalas, line 69: Same with this line, well maybe the ones that are too long in page controllers I'll refactor when I'll go through them.

src/Pages/ChangePassword.cshtml.cs
```diff
@@ -66,17 +70,45 @@ public async Task<IActionResult> OnGet()
...
+                var currencyCodeList = currencyList.Select(currency => currency.Code).ToList();
+
+                var options = CacheKeys.DefaultCurrencyCacheOptions();
+
+                Cache.Set(CacheKeys.CurrencyList, currencyCodeList, options);
+                CurrencyOptions.AddRange(currencyCodeList);
+
+                //Also set user currency 
+                var userCurrency = currencyList.First(x => x.Code == UserModel.Currency);
+                Cache.Set(CacheKeys.UserCurrency, userCurrency, options);
```
Stokalas, line 112: So this is the same as previous reading of currencies, just it stores data in user cache and reads only if its empty or outdated data?

src/Utilities/CacheKeys.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using Microsoft.Extensions.Caching.Memory;
+
+namespace ePiggyWeb.Utilities
+{
+    public static class CacheKeys
```
Stokalas, line 6: Could you explain what this class does?  in short

Stokalas (commented):
Since most of code in controllers are repeating I tried to ask questions about unique code. Did not test, so no clue if it works, questions for now 🙂

src/Pages/ChangePassword.cshtml.cs
ciliakas: That's pretty much exactly what it does :D

src/Utilities/CacheKeys.cs
ciliakas: It's only for more convenient use of the cache - the cache keys are saved there so it's easier to change them all in bulk if needed, and I put a method for default options for currency because they're always the same

PR #66, opened 2020-12-18
Cache alerts

Stokalas:
Added alerts when error in reading currency data.


ciliakas (commented):
Tested this out, works as intended, but I think that you forgot to implement the alert in saving suggestions :D

PR #68, opened 2020-12-18
Added pagination to the back end

ciliakas:
Created a method for pagination completely in the back, it only returns the entries needed to display and the amount of pages there are in total.
Can't merge automatically because I also implemented this pagination in expenses as an example :D


src/DataBase/EntryDatabase.cs
```diff
@@ -175,35 +171,27 @@ public async Task<IEntry> ReadAsync(Expression<Func<IEntryModel, bool>> filter,
...
-                    Title = row.Field<string>("Title"),
-                    UserId = row.Field<int>("UserId")
-                } as IEntry).ToList();
-
-            list.AddRange(items);
-            return list;
+            var list = new EntryList(entryType);
+            list.AddRange(allEntries.Skip((pageNumber - 1) * pageSize).Take(pageSize));
+            var numberOfPages = (int) Math.Ceiling(decimal.Divide(allEntries.Count, pageSize));
+            return new Tuple<IEntryList, int>(list, numberOfPages);
```
Stokalas, line 194: Really thoughtful to return number of pages :)

src/Pages/Expenses.cshtml.cs
```diff
@@ -138,15 +139,13 @@ private async Task SetData()
...
-                    .GetPage(CurrentPage, PageSize);
-
-                AllExpenses = Expenses.GetSum();//Replaced with sum from database.
+                var entryList = await EntryDatabase.ReadListAsync(UserId, EntryType.Expense);
+                Expenses = entryList.GetFrom(StartDate).GetTo(EndDate);
+                var (expenseList, totalPages) = await EntryDatabase.ReadByPage(x => x.Date >= StartDate && x.Date <= EndDate, UserId,
+                    EntryType.Expense, CurrentPage, PageSize);
+                ExpensesToDisplay = expenseList;
+                TotalPages = totalPages;
+                AllExpenses = Expenses.GetSum();
```
Stokalas, line 148: So this is the part where it collides, we have to read whole list to get sum, and then we read just a part of entries to display? :D

Stokalas (approved):
Really like the changes, and will implement I'll implement this to income, just we need to decide what to do in parts where it collides.

src/Pages/Expenses.cshtml.cs
ciliakas: Yeah, after I created this I realised that there is not much to this method... :D Because we read all the entries anyways to get the sum, with implementation like this it's actually a lot slower then it was before :D

Stokalas: But there's has to be a workaround, since everytime to read 3000 entries to display 10 is illogical, maybe the sum then should be removed from display and we won't need it?

ciliakas: We can do that, but then there has to be a place where user can see all this data :D

PR #69, opened 2020-12-19
Order by implementation

DeividasLT5:
Small requirement implementation, shouldn't change how program functions


ciliakas (approved):
Almost forgot this, definitely needed this for the requirements

src/DataManagement/Saving/AlternativeSavingCalculator.cs
```diff
@@ -33,9 +34,15 @@ public CalculationResults GetSuggestedExpensesOffers(IEntryList entryList, IGoal
                 generateRandomData = true;              
             }
 
-            for (var i = enumCount; i > (int)Importance.Necessary; i--)
+            var listToUse = generateRandomData ? expensesRandomList : entryList;
+            var groupedByImportance = (from entry in listToUse
+                                       where entry.Importance != 1
+                                       group entry by entry.Importance).ToArray();
```
Stokalas, line 40: Could you explain what does it return? (since we need to use `ToArray()` )

Stokalas (approved):
It needed this PR long ago :D Nice, left one question.

src/DataManagement/Saving/AlternativeSavingCalculator.cs
ciliakas: It returns an IEnumerable of IGrouping, but since it is more efficient to traverse an array than an IEnumerable, we cast it to an array. Of course this is not really necessary, but since originally we thought maybe the list will be traversed multiple times this bit was added ( as suggested by resharper) :D

PR #70, opened 2020-12-20
More fixes!

Stokalas:
- Changed pagination so that it always display: First page btn, prev page btn, 5 numbered buttons, next page btn, last page btn.
- All filters are now limited from 2016-01 till today.
- Saving suggestions only allows to filter by month (since it suggests how many months you would need and 1 year period shows that you would save up faster than when 1 month entered)
- Bug fixes


src/Pages/Expenses.cshtml.cs
```diff
@@ -132,6 +132,10 @@ public async Task<IActionResult> OnPostNewEntry()
         public async Task<IActionResult> OnPostDelete()
         {
             var selected = Request.Form["chkEntry"].ToString();
+            if (string.IsNullOrEmpty(selected))
+            {
+                return RedirectToPage("/income");
```
ciliakas: This redirects to income rather than expenses

ciliakas (changes requested):
To start with, pagination seems to work really well, however I found some bugs elsewhere and have a couple of suggestions.

I think that a better way to handle the delete button when nothing is selected is to use javascript, I will soon post my version of it as a small PR.

bug1

You can press the little x button on the calendar, and then it resets to yyyy-mm-dd, and then you can still click ok, which sends you to year 0001. My solution to this would bet to redirect to current date or to the saved filter date

bug2

I think we're moving in the right direction with limited the amount of choices of filtering in suggestions, but I do think that using a text box like this is very prone to user errors. We'd need quite good regex to catch everything, especially when considering that every user is using different locales and so on. My suggestion would be to just set a few dates as buttons, and let the user select between those. I would use month, quarter, half year and year buttons. This removes all user error and these dates correlate quite well with usual financial timelines.

Stokalas:
I don't think I can agree with "I would use month, quarter, half year and year buttons." Since calculations are base on months, and for example if we calculate using 1 month's data it would show that you need to save for 6 months, but then when used 6 months, more entries, reaches sum faster and it shows that you need to save for 1 month, which is illogical since these all entries were achieved during 6 months not one month. I hope I explained it clearly. So I think that month's range is the only viable option.

ciliakas:
> 
> 
> I don't think I can agree with "I would use month, quarter, half year and year buttons." Since calculations are base on months, and for example if we calculate using 1 month's data it would show that you need to save for 6 months, but then when used 6 months, more entries, reaches sum faster and it shows that you need to save for 1 month, which is illogical since these all entries were achieved during 6 months not one month. I hope I explained it clearly. So I think that month's range is the only viable option.

As far as I'm aware, the calculations themselves never take into account months or any sort of dates at all, it just works with the entry list given (the name called 'monthlySuggestions' is misleading - it is suggestions based on importance actually). The calculator works with averages, so if I'm not mistaken, the amount of entries given or the dates from when they start are not that important to the calculation itself

Stokalas:
> 
> 
> > I don't think I can agree with "I would use month, quarter, half year and year buttons." Since calculations are base on months, and for example if we calculate using 1 month's data it would show that you need to save for 6 months, but then when used 6 months, more entries, reaches sum faster and it shows that you need to save for 1 month, which is illogical since these all entries were achieved during 6 months not one month. I hope I explained it clearly. So I think that month's range is the only viable option.
> 
> As far as I'm aware, the calculations themselves never take into account months or any sort of dates at all, it just works with the entry list given (the name called 'monthlySuggestions' is misleading - it is suggestions based on importance actually). The calculator works with averages, so if I'm not mistaken, the amount of entries given or the dates from when they start are not that important to the calculation itself

I don't think you understood what I'm trying to say. Just look at this from this perspective:
1) If data from one month used - displays: 5 months to save
2) If data from six months used - displays: 1 month to save.
And this is really illogical. And this happens because of how calculations work - it iterates through all entries lowers them and checks if the sum after lowering reaches goal price. So if we give bigger entry list it will collect sum faster and display that less months required to save the sum. But then it seems like these changes for entries (good example recurring ones) should be made in one month to save, eventhough they're from whole year lets say :D Just try to let that sink in.

ciliakas:
> 
> 
> > > I don't think I can agree with "I would use month, quarter, half year and year buttons." Since calculations are base on months, and for example if we calculate using 1 month's data it would show that you need to save for 6 months, but then when used 6 months, more entries, reaches sum faster and it shows that you need to save for 1 month, which is illogical since these all entries were achieved during 6 months not one month. I hope I explained it clearly. So I think that month's range is the only viable option.
> > 
> > 
> > As far as I'm aware, the calculations themselves never take into account months or any sort of dates at all, it just works with the entry list given (the name called 'monthlySuggestions' is misleading - it is suggestions based on importance actually). The calculator works with averages, so if I'm not mistaken, the amount of entries given or the dates from when they start are not that important to the calculation itself
> 
> I don't think you understood what I'm trying to say. Just look at this from this perspective:
> 
>     1. If data from one month used - displays: 5 months to save
> 
>     2. If data from six months used - displays: 1 month to save.
>        And this is really illogical. And this happens because of how calculations work - it iterates through all entries lowers them and checks if the sum after lowering reaches goal price. So if we give bigger entry list it will collect sum faster and display that less months required to save the sum. But then it seems like these changes for entries (good example recurring ones) should be made in one month to save, eventhough they're from whole year lets say :D Just try to let that sink in.

Yeah, I think I understand what you mean now. If we're comparing suggestions, they all have to be of the same length of time.

Stokalas:
In short calculations return how many times all given entries should be repeated with reduced amount to achieve goal. Since we display it as "months" the other time periods break the logic.

ciliakas (approved):
After having a discussion with you about the logic behind this kind of filtering, I agree that this is the right way to do the filtering for suggestions.

PR #71, opened 2020-12-21
Disable delete button (Javascript)

ciliakas:
So this turned out to be more that I originally planed :D 
Added disabling of filter and delete buttons when conditions are not met to use them ( date is out filtering range, no entries selected to delete)
Also added everything into a single js file and included it in the shared layout


Stokalas (approved):
Really nice touch on disabling that button, and I think we can agree to keep validation of button click  when 0 selected just in case :D

ciliakas:
> 
> 
> Really nice touch on disabling that button, and I think we can agree to keep validation of button click when 0 selected just in case :D

Fully agree, let's keep all the checks just in case the javascript is broken somehow

src/Pages/Shared/_Layout.cshtml
```diff
@@ -11,117 +11,121 @@
...
         document.addEventListener("DOMContentLoaded",
             function() {
                 particlesJS.load('particles-js', 'js/particle-cfg.json');
             });
     };
-    </script>
+        </script>
+
+
+        <script src="~/js/site.js"></script>
```
Stokalas: Very minor detail, but I think 2 empty lines are unnecessary, 1 would be enough :D

src/wwwroot/js/site.js
```diff
@@ -2,3 +2,41 @@
 // for details on configuring this project to bundle and minify static web assets.
```
Stokalas, line 2: Really nice that you moved javascript into separate file since now we don't have to repeat same code and overall better structure.

Stokalas (approved):

src/Pages/Shared/_Layout.cshtml
ciliakas: Agree, did this by mistake, so in new update I also removed some old commented code while I had the chance :D

PR #74, opened 2020-12-21
Full Currency API Implementation 

ciliakas:
This is the **big** currency API update, which fully integrates the API into our program. While this is a big pull request, some things do not have to be reviewed because they were already reviewed previously, or are simple to review because they are analogous between many files.
We should check everything well, since this is the last feature update.

Things that already have been reviewed previously:
- Javascript changes
- Pagination changes
- Exception handling for the API
- API caching


src/DataManagement/Entries/EntryList.cs
```diff
@@ -109,7 +109,7 @@ public static IEntryList RandomList(IConfiguration configuration, EntryType entr
             {
                 var amount = random.Next(50, 500);
                 var importance = random.Next(minImportance, maxImportance);
-                IEntry entry = Entry.CreateLocalEntry(thing.Value, amount, DateTime.UtcNow, false, importance);
+                IEntry entry = Entry.CreateLocalEntry(thing.Value, amount, DateTime.UtcNow, false, importance, "EUR");
```
Stokalas: Named arguments could be used for boolean and "EUR" for clarification.

src/DataManagement/Goals/GoalList.cs
```diff
@@ -28,7 +28,7 @@ public static IGoalList RandomList(IConfiguration configuration)
             {
                 var sec = thing.GetChildren();
                 var configurationSections = sec as IConfigurationSection[] ?? sec.ToArray();
-                IGoal goal = Goal.CreateLocalGoal(configurationSections.Last().Value, decimal.Parse(configurationSections.First().Value));
+                IGoal goal = Goal.CreateLocalGoal(configurationSections.Last().Value, decimal.Parse(configurationSections.First().Value), "EUR");
```
Stokalas: Same with name argument for "EUR"

src/CurrencyAPI/CurrencyConverter.cs
```diff
@@ -1,82 +1,138 @@
...
+            {
+                return goalList;
+            }
+
+            var (currencyList, exception2) = await GetCurrencyList(userId);
+            if (exception2 != null)
+            {
+                throw exception2;
+            }
+
```
Stokalas: a bit weird naming "exception2"

src/Pages/ChangePassword.cshtml.cs
```diff
@@ -43,19 +44,19 @@ public class ChangePasswordModel : PageModel
...
         private EmailSender EmailSender { get; }
-        private CurrencyConverter CurrencyConverter { get; }
-        public UserModel UserModel { get; private set; }
+        public string UserCurrencyCode { get; set; }
         private IMemoryCache Cache { get; }
         public bool FailedToGetCurrencyList { get; set; }
+        private CurrencyConverter CurrencyConverter { get; }
 
-        public ChangePasswordModel(PiggyDbContext piggyDbContext, IOptions<EmailSender> emailSenderSettings, ILogger<ChangePasswordModel> logger, CurrencyConverter currencyConverter, IMemoryCache cache)
+        public ChangePasswordModel(PiggyDbContext piggyDbContext, IOptions<EmailSender> emailSenderSettings, ILogger<ChangePasswordModel> logger, IMemoryCache cache, CurrencyConverter currencyConverter)
```
Stokalas: Forgotten long line

src/Pages/ChangePassword.cshtml.cs
```diff
@@ -65,99 +66,101 @@ public async Task<IActionResult> OnGet()
...
+                    case HttpRequestException ex:
+                        _logger.LogInformation(ex.ToString());
+                        ErrorMessage = "Failed to load currency information!";
+                        break;
+                    default:
+                        // Failed to even get the user currency from the database, the big one
+                        _logger.LogInformation(exception.ToString());
+                        ErrorMessage = "Failed to connect to database!";
+                        break;
                 }
```
Stokalas: I think property `FailedToGetCurrencyList` should be set to true instead of `ErrorMessage`?
Maybe I myself missed this, just noting it to remember

src/Pages/Entry.cshtml.cs
```diff
@@ -31,7 +31,7 @@ public EntryModel(HttpClient httpClient, IConfiguration configuration, IMemoryCa
 
         public async Task OnGet()
         {
-            var currencyConverter = new CurrencyConverter(HttpClient, Configuration);
+            var currencyConverter = new CurrencyApiAgent(HttpClient, Configuration);
             var list = await currencyConverter.GetList();
             var sb = new StringBuilder();
```
Stokalas: Maybe now is the time to remove this testing files "Entry"?

src/Pages/Expenses.cshtml
```diff
@@ -137,6 +139,7 @@ else
                                         </div>
                                         <div class="col-12 col-md-5 col-lg-4">
                                             <span class="text-info">Amount:</span>
+                                            @*@NumberFormatter.FormatCurrency(item.Amount, Model.CurrencyException? item.Currency : Model.CurrencySymbol)*@
```
Stokalas: This line was forgotten? :D

src/Pages/Income.cshtml
```diff
@@ -83,17 +83,22 @@ else
         <input type="hidden" name="handler" value="Filter" />
         <label class="text-light col-md-4">Period:</label>
         <div class="row justify-content-center">
-            <input class="col-md-3 col-sm-5 col-5" for="startDate" asp-for="StartDate" type="date" value="@Model.StartDate.ToShortDateString()" />
+            <input class="col-md-3 col-sm-5 col-5" onchange="handleFilter(this)"  for="startDate" min="2016-01-01" max="@DateTime.Now.AddMonths(1).ToShortDateString()" asp-for="StartDate" type="date" value="@Model.StartDate.ToShortDateString()" />
             <label class="text-light ml-2 mr-2"> - </label>
-            <input class="col-md-3 col-sm-5 col-5" for="endDate" asp-for="EndDate" type="date" value="@Model.EndDate.ToShortDateString()" />
+            <input class="col-md-3 col-sm-5 col-5" onchange="handleFilter(this)"  for="endDate" min="2016-01-01" max="@DateTime.Now.AddMonths(1).ToShortDateString()" asp-for="EndDate" type="date" value="@Model.EndDate.ToShortDateString()" />
```
Stokalas: Note for myself: since DateTime.Now is called several times and it is costly operation, property inside controller with DateTime.Today should be created.

Stokalas (changes requested):
Well this seems a big one, but a lot of changes are just moving of HTML text :D Left quite a few comments and notices.

src/Pages/Income.cshtml
Stokalas: Maybe I'll do this in separate branch after merging this one, not to create more chaos 😆

src/Pages/ChangePassword.cshtml.cs
ciliakas: The property FailedToGetCurrencyList  is not used anywhere in the code though

src/DataBase/EntryDatabase.cs
```diff
@@ -140,28 +130,26 @@ public async Task<bool> DeleteListAsync(Expression<Func<IEntryModel, bool>> filt
             //var predicate = Expression<Func<IEntryModel, bool>>.Create();
```
MatasLa, line 123: Not sure if this is still needed, if no could be removed.

src/DataManagement/Goals/GoalList.cs
```diff
@@ -14,21 +15,22 @@ public override string ToString()
             foreach (var goal in this)
             {
                 sb.Append(goal);
-                //sb.Append("\n");
```
MatasLa, line 17: Again another commented line, could be removed if not planning to use.

src/Startup.cs
```diff
@@ -45,6 +45,7 @@ public void ConfigureServices(IServiceCollection services)
             services.AddScoped<IGoalDatabase, GoalDatabaseSql>();
```
MatasLa, line 45: I guess lets leave it as GoalDatabase as default.

src/Pages/ForgotPassword.cshtml.cs
```diff
@@ -81,16 +80,12 @@ public async Task<IActionResult> OnPost()
                     Response.Cookies.Delete("Email");
                     return RedirectToPage("/login");
                 }
-                else
-                {
-                    ErrorMessage = "Password did not match!";
-                }
+                ErrorMessage = "Password did not match!";
```
MatasLa: "Passwords did not match!" would be better

MatasLa (changes requested):
A lot of work done, but noticed while testing that duplicate entries are added. Need to fine the problem and fix it. Everything except that seems fine.

src/Startup.cs
ciliakas: We will do that, but now it was important to test the GoalDatabaseSQL as well, so that's why I left it in

src/Pages/Expenses.cshtml
```diff
@@ -139,7 +139,6 @@ else
                                         </div>
                                         <div class="col-12 col-md-5 col-lg-4">
                                             <span class="text-info">Amount:</span>
-                                            @*@NumberFormatter.FormatCurrency(item.Amount, Model.CurrencyException? item.Currency : Model.CurrencySymbol)*@
                                             @NumberFormatter.FormatCurrency(item.Amount, Model.CurrencySymbol)
```
Stokalas: So we don't need to use that commented line? I remember you've put it here to use later?

Stokalas (approved):
I'm not sure if github not showing me removal of comments Matas mentioned, but other than that it seems ok.

src/Pages/Expenses.cshtml
ciliakas: I completely forgot about this :D I'll go and actually check out how this works and see if we need to keep this or not

src/Pages/Income.cshtml
```diff
@@ -143,7 +143,7 @@ else
                                         </div>
                                         <div class="col-12 col-md-5 col-lg-4">
                                             <span class="text-info">Amount:</span>
-                                            @NumberFormatter.FormatCurrency(item.Amount, Model.CurrencySymbol)
+                                            @NumberFormatter.FormatCurrency(item.Amount, Model.CurrencyException? item.Currency : Model.CurrencySymbol)
```
Stokalas, line 146: How exactly did this alter the display when currency is down? And maybe this should be used in for example Saving suggestions, since there are amounts of entries as well.

ciliakas: So this works by checking if there was an currency exception, if there was - it displays the currency code of the entry itself, otherwise, in the normal situation, it will display in the currency the user has currently selected

ciliakas: But if I'm not mistaken, the calculator will not run, if we can't convert the currencies, so there isn't really any point to doing any changes since they won't be displayed anyway

Stokalas: You're right on this one

PR #75, opened 2020-12-22
Datetime optimization

Stokalas:
This is small optimization, where front pages get access to Today property instead of calling DateTime.Today.


ciliakas (changes requested):
Good change, but I think that the set methods to these properties should be made private, since now we're trying for the most clean code possible :D

ciliakas (approved):
Approving since fixed the setting issue, and big improvement in saving suggestions controller

PR #76, opened 2020-12-22
Refactoring

ciliakas:
Mainly deleted unnecessary code, renamed a few files, and implemented DI for monthly report calculator


src/Utilities/EmailSender.cs
```diff
@@ -2,24 +2,25 @@
 using System.Net;
 using System.Net.Mail;
 using System.Threading.Tasks;
-using ePiggyWeb.DataBase;
+// ReSharper disable MemberCanBePrivate.Global
+// ReSharper disable UnassignedGetOnlyAutoProperty
```
Stokalas, line 6: Are these necessary, because if resharper is not used, a bit weird ?

src/Program.cs
```diff
@@ -3,14 +3,14 @@
 
 namespace ePiggyWeb
 {
-    public class Program
+    public static class Program
```
Stokalas, line 6: Why was this made static?

src/DataManagement/MonthlyReport/MonthlyReportCalculator.cs
```diff
@@ -26,19 +26,18 @@ public class MonthlyReportCalculator
 
         private DateTime StartTime { get; set; }
         private DateTime EndTime { get; set; }
-        public CurrencyConverter CurrencyConverter { get; set; }
+        private CurrencyConverter CurrencyConverter { get; set; }
 
-        public MonthlyReportCalculator(IGoalDatabase goalDatabase, EntryDatabase entryDatabase, int userId,
-            CurrencyConverter currencyConverter)
+        public MonthlyReportCalculator(IGoalDatabase goalDatabase, EntryDatabase entryDatabase, CurrencyConverter currencyConverter)
```
Stokalas, line 31: So when this itself is included in Startup it has access to other DI objects?

Stokalas (approved):
Since most of the changes was deletion, nothing to comment much, but still very needed work. Nice.

src/Utilities/EmailSender.cs
ciliakas: It looks really bad without them in resharper though, because all the properties are marked as being bad, because reshaper doesn't catch that these properties are binded through configuration. And since probably we will be showing our project with resharper, I think it's worth keeping them in

src/Program.cs
ciliakas: Resharper suggestion :D

src/DataManagement/MonthlyReport/MonthlyReportCalculator.cs
ciliakas: Yes, so we don't have to create this manually in the controller. However I did have to change the calculate method to pass the user id parameter there

PR #77, opened 2020-12-22
Calculation Update

ciliakas:
Changed the calculator to include the average income in the calculation


src/DataManagement/Saving/SavingCalculator.cs
```diff
@@ -4,113 +4,144 @@
...
-                            break;
-                        default:
-                            throw new Exception("Unexpected saving type");
-                    }
-                    entrySuggestions.Add(new SavingSuggestion(entry, amountAfterSaving));
-
-                    sumsOfAmountByImportanceAdjusted[i - 1] += amountAfterSaving;
-                    sumsOfAmountByImportanceDefault[i - 1] += entry.Amount;
-                    entryAmounts[i - 1]++;
+                    months++;
```
Stokalas: Maybe this could be calculated by just Newest - Oldest (by converting years to months and adding months) ?

src/DataManagement/Saving/SavingCalculator.cs
```diff
@@ -4,113 +4,144 @@
...
+                };
+
+                SavingSuggestionByImportanceList.Add(new SavingSuggestionByImportance(newAverage, importanceAverage, (Importance)i));
             }
-            for (var i = enumCount; i > (int)Importance.Necessary; i--)
+        }
+
+        private void CalculateSavingTime()
+        {
+            var increasedMonthlySavings = SavingSuggestionByImportanceList.Sum(savingSuggestionByImportance => savingSuggestionByImportance.OldAverage - savingSuggestionByImportance.NewAverage);
```
Stokalas: A bit too long line

src/DataManagement/Saving/SavingCalculator.cs
```diff
@@ -4,113 +4,144 @@
...
+            var amountToSave = Goal.Amount - StartingBalance;
+            MonthsToSave = (int)Math.Ceiling(decimal.Divide(amountToSave, MonthlyIncome + increasedMonthlySavings));
+            if (MonthsToSave > 1000)
```
Stokalas: I think 1000 months is overkill and max time to save should be reduced to, let's say 5 years? (60 months)

src/DataManagement/Saving/SavingCalculator.cs
```diff
@@ -4,113 +4,144 @@
...
+        public SavingCalculator(IEntryList expenseList, IEntryList incomeList, IGoal goal, decimal startingBalance)
         {
-            var entrySuggestions = new List<ISavingSuggestion>();
-            var monthlySuggestions = new List<SavingSuggestionByImportance>();
-            var expensesRandomList = EntryList.RandomList(configuration, EntryType.Expense);
-            var generateRandomData = false;
+            // Reikia gauti average income
+            // Tada reikia, kad kas menesi prisidetu recurring + average per menesi
+            // I menesi sutaupai = averageIncome + savedInAMonth
+            // savedInAMonth =
```
Stokalas: Are you going to implement this yourself or this was meant for somebody else?

Stokalas (changes requested):
Requesting changes simply for that comments part.

src/DataManagement/Saving/SavingCalculator.cs
ciliakas: I'd say something like 20 years would be at the upper part of reasonable, what do you think?

src/DataManagement/Saving/SavingCalculator.cs
ciliakas: Left this here by mistake

src/DataManagement/Saving/SavingCalculator.cs
ciliakas: Yeah this is a silly way to do it :D I'll fix this

src/DataManagement/Saving/SavingCalculator.cs
Stokalas: Maybe something like if item is bellow 10k price - 20 years, bellow 5 years? But this creates more problems as different value in different currencies, so maybe middle ~12 years? :D

src/DataManagement/Saving/SavingCalculator.cs
```diff
@@ -78,21 +73,28 @@ private void CalculateMonthlyIncome()
 
         private void GetAverages()
         {
-            for (var i = EnumCount; i > (int)Importance.Necessary; i--)
+            var groupedByImportance =
+                from entry in ExpenseList
+                where entry.Importance != 1
+                group entry by entry.Importance;
```
Stokalas, line 66: Really nice that  you've remembered this :D

Stokalas (approved):

src/DataManagement/Saving/SavingCalculator.cs
ciliakas: I just set to 10 years for simplicity :D

